PR #2, opened 2019-10-29
Added instructions for chromium.

xvzcf:


dstebila:
Is it possible to create a Docker image that does this build and/or contains a pre-built version of Chromium?  (Although perhaps that shows my ignorance of Docker -- can you run a GUI application from inside a container?)  It would be nice of we could provide prebuilt Docker images of Chromium, nginx, etc. which people could use together.

xvzcf:
We can discuss that [here](https://github.com/open-quantum-safe/oqs-software/issues/3).

PR #4, opened 2019-10-30
Added instructions for httpd.

xvzcf:


httpd/Dockerfile
```diff
@@ -0,0 +1,77 @@
...
+                          libexpat1-dev \
+                          libpcre3-dev \
+                          # misc
+                          wget;
+
+# get sources
+WORKDIR /opt
+RUN git clone --single-branch --branch master https://github.com/open-quantum-safe/liboqs && \
+    git clone --single-branch --branch OQS-OpenSSL_1_1_1-stable https://github.com/open-quantum-safe/openssl ossl-src && \
+    wget http://apache.forsale.plus/httpd/httpd-${HTTPD_VERSION}.tar.gz && tar -zxvf httpd-${HTTPD_VERSION}.tar.gz;
```
dstebila: The URL here is apache.forsale.plus?  That seems really suspicious.

xvzcf: It does sound suspicious, but it is an official apache mirror as per [this list](https://www.apache.org/mirrors/). We could change it, [this stackoverflow answer](https://stackoverflow.com/questions/38592289/downloading-file-from-the-right-apache-mirror-with-wget) looks interesting.

dstebila: Yeah, maybe take the approach in the StackOverflow answer. Or if you want to have a hard-coded URL, then maybe use one of the top Apache ones (www-us.apache.org).

PR #6, opened 2019-12-16
Added OQS centos development image.

xvzcf:
As per discussion with @baentsch, this is a draft PR adding a docker image which provides a development environment with liboqs and OQS-OpenSSL installed in system locations. Compilation of headers and dynamic loading can be done out of the box, and dynamic linking requires only the `-lssl -lcrypto -loqs` flags.

I'll add a README and refactor next.

### DIFF
--- a/development/centos/Dockerfile
+++ b/development/centos/Dockerfile
@@ -1,0 +1,38 @@
+FROM centos:7
+
+RUN yum -y update && yum -y install gcc git autoconf automake libtool perl-core zlib-devel make yum-utils wget unzip doxygen graphviz xsltproc python3 python3-pip  && yum check-update
+
+RUN pip3 install -U pytest nose rednose pytest-xdist
+
+# liboqs requires better gcc than default 4.8:
+RUN yum -y remove gcc
+RUN yum -y install centos-release-scl
+RUN yum -y install devtoolset-8-gcc*
+RUN yum -y install libtool
+RUN echo "source /opt/rh/devtoolset-8/enable" >> /root/.bashrc
+
+WORKDIR /root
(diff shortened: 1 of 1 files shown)


baentsch:
I have not been asked for review, but I had time anyway :-)

- I like this image much better than the openssl draft one I just commented on: git is in sync when logging in, the image is still a bit large but much smaller than the Ubuntu one (which is weird, but good, given how large Centos normally is).
- I ran into the same performance limitations on a single-core CPU. For a possible solution, please take a look at [this branch example](https://github.com/baentsch/testing/blob/limit-makej/integration/oqs_openssl/scripts/build_openssl.sh) : I introduced an environment variable, which, when **not** set throttles building to `make -j2`. For people that *read* the documentation (and that know what they do) this limit can simply be removed by setting the variable. I'd think this approach is safer than what the build{liboqs|openssl} scripts currently do (aggressive/unlimited `make -j` -- which apparently also is/was a problem for CI)
- When using your image to build curl, it worked like a charm without any need to set CPP- or LDFLAGS: Nice (&better than my proposal -- although that one is still several hundred MB smaller :-) Maybe worthwhile having a chat as to what matters more (?). 

--> What about agreeing on the goals for the Docker images first? I took a shot at this with [this Wiki](https://github.com/baentsch/testing/wiki/Docker-image-goals) as I couldn't create a page in OQS/testing: Please place it somewhere where you deem it helpful and everyone can chime in and we come to agreement before continuing our work.

dstebila:
Is this PR still being worked on?  Or is this obsolete and to be replaced with something new?

xvzcf:
It's not obsolete, I will continue to work on it.

xvzcf:
Closing for now, might re-open in OQS-OpenSSL.

PR #7, opened 2020-02-04
API update, size optimization, auto-start

baentsch:
This updates the curl build
- to the standard openssl APIs
- creating run-time SIG_ALG, KEM_ALG parameterisation
- reducing its size considerably
- eliminating the need for '--insecure' communication
- automating its use
- enhancing the documentation
- enabling performance testing [addressing an OQS-openssl issue](https://github.com/open-quantum-safe/openssl/issues/10)
- making it easier to deploy in "secured" K8S clusters (non-root user)

### DIFF
--- a/curl/Dockerfile
+++ b/curl/Dockerfile
@@ -1,7 +1,11 @@
-FROM debian:buster-slim
+# Multi-stage build: First the full builder image:
 
+FROM debian:buster-slim as intermediate
+
+# define the Curl version to be baked in
 ENV CURL_VERSION 7.66.0
 
+# Get all software packages required for builing all components:
 RUN apt-get update -qq \
     && apt-get install -y build-essential \
                           git \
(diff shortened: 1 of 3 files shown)


baentsch:
Would it make sense to push this onto docker hub (e.g., openquantumsafe/curl) -- from a 'conservatively' (regarding CPU-flags) building platform, maybe CircleCI directly? Could constitute new support obligations as this makes things trivial to (mis)use by ..... "specialists". But also could help questions asking for help building the system (along the lines, "if you can't build stuff, use this image"...

dstebila:
> Would it make sense to push this onto docker hub (e.g., openquantumsafe/curl) -- from a 'conservatively' (regarding CPU-flags) building platform, maybe CircleCI directly? Could constitute new support obligations as this makes things trivial to (mis)use by ..... "specialists". But also could help questions asking for help building the system (along the lines, "if you can't build stuff, use this image"...

Yes.  As for pushing from CircleCI, it seems we get a variety of CPUs when using CircleCI, sometimes AVX512 sometimes not, so that might not be the best approach.  If one of us has an older machine (e.g., AVX2, AES-NI, but not much more) maybe we could use that?

PR #10, opened 2020-03-02
Build type & Performance testing documented

baentsch:
Please check whether this PR could also serve to close OQS-OpenSSL issues [#10](https://github.com/open-quantum-safe/openssl/issues/10) and [#35](https://github.com/open-quantum-safe/openssl/issues/35).

Sample documentation for resulting generic Docker image at [Docker Hub](https://hub.docker.com/repository/docker/openqsafe/curl)

### DIFF
--- a/curl/Dockerfile
+++ b/curl/Dockerfile
@@ -20,6 +20,9 @@ ENV CURL_VERSION 7.66.0
 # Default location where all binaries wind up:
 ARG INSTALLDIR=/opt/oqssa
 
+# liboqs build type variant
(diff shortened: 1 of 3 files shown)


dstebila:
Very nice!

I'd say this does resolve open-quantum-safe/openssl#10, but it would be good to add a note about this in the OQS-OpenSSL README.md so people know about it.

This doesn't resolve open-quantum-safe/openssl#35, as that would involve speed testing of the algorithm standalone, not in the context of TLS.

PR #12, opened 2020-03-11
docker image auto-build-test-push cycle added

baentsch:
This adds CI testing to oqs-demos as well as automatic docker image creation on successful test completion. The latter only works if the CCI context `openquantumsafe` exists and is loaded with the environment variables DOCKER_LOGIN and DOCKER_PASSWORD. The latter can be a [docker hub access token](https://docs.docker.com/docker-hub/access-tokens). The images will be re-build whenever this repo,`liboqs/master`, or `openssl/OQS-OpenSSL_1_1_1-stable` change.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -36,7 +36,7 @@ localCheckout: &localCheckout
         command: docker tag oqs-curl $DOCKER_LOGIN/curl-generic && docker push $DOCKER_LOGIN/curl-generic && docker tag oqs-httpd-img $DOCKER_LOGIN/httpd-generic && docker push $DOCKER_LOGIN/httpd-generic && docker tag oqs-nginx-img $DOCKER_LOGIN/nginx-generic && docker push $DOCKER_LOGIN/nginx-generic
 
 
-# As of now (Mar 2020) CircleCI does not support building current OSX docker images... :-(
+# Reactivate when issue #13 is resolved:
 #.mac_job: &macjob
 #    macos:
 #       xcode: "11.3.0"
@@ -80,7 +80,7 @@ workflows:
   version: 2
   build:
     jobs:
-      - debian-buster-amd64:
+      - ubuntu-bionic-x86_64:
           context: openquantumsafe
   nightly:
     triggers:
@@ -93,6 +93,6 @@ workflows:
     jobs:
 #      - macOS:   
 #          context: openquantumsafe
-      - ubuntu-bionic-x86_64:
+      - debian-buster-amd64:
           context: openquantumsafe
 
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,98 @@
+version: 2
+
+# CircleCI doesn't handle large file sets properly for local builds
+# https://github.com/CircleCI-Public/circleci-cli/issues/281#issuecomment-472808051
+localCheckout: &localCheckout
+  run: |-
+    PROJECT_PATH=$(cd ${CIRCLE_WORKING_DIRECTORY}; pwd)
+    mkdir -p ${PROJECT_PATH}
+    cd /tmp/_circleci_local_build_repo
+    git ls-files -z | xargs -0 -s 2090860 tar -c | tar -x -C ${PROJECT_PATH} 
+    cp -a /tmp/_circleci_local_build_repo/.git ${PROJECT_PATH}
+
+.linux_job: &linuxjob
+  docker:
+    - image: ${IMAGE}
+  steps:
+    - checkout # change this from "checkout" to "*localCheckout" when running CircleCI locally
+    - setup_remote_docker
+    - run:
+        name: docker login
+        command: echo $DOCKER_PASSWORD | docker login --username $DOCKER_LOGIN --password-stdin
+    - run:
+        name: checkout dependencies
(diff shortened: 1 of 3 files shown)


xvzcf:
It seems the `docker` command is not available in the `openquantumsafe/ci-debian-buster-amd64` image, which causes the `debian-buster-amd64` job to fail. Looking into it.

baentsch:
That's weird - I just pulled `openquantumsafe/ci-debian-buster-amd64` and ran `docker` OK. `docker ps` fails as the daemon is not running but I assumed (and thought tested) that this is OK when the docker engine is provided by the (CircleCI) host -- but I may be wrong: As it's my "upgrade" I shall fix it: @xvzcf : Don't waste too much time.

xvzcf:
We haven't set this up in CircleCI yet. Where is the build being triggered?

baentsch:
I just triggered this build on my CircleCI account to see whether errors occur & it just completed OK : https://app.circleci.com/pipelines/github/baentsch/oqs-software/22/workflows/f910c251-5e0d-472b-b380-37e6194a5c20/jobs/40 (let me know if you cannot follow the link & let's quickly talk in 5 mins as to what's not looking OK).

xvzcf:
> That's weird - I just pulled `openquantumsafe/ci-debian-buster-amd64` and ran `docker` OK. `docker ps` fails as the daemon is not running but I assumed (and thought tested) that this is OK when the docker engine is provided by the (CircleCI) host -- but I may be wrong: As it's my "upgrade" I shall fix it: @xvzcf : Don't waste too much time.

Sorry, false alarm. It turns out my cache was just retrieving an earlier version of the CI image.

.circleci/config.yml
```diff
@@ -0,0 +1,98 @@
...
+        command: .circleci/git_no_checkin_in_last_day.sh || (cd httpd && docker build -t oqs-httpd-img . && docker run --detach --rm --name oqs-httpd oqs-httpd-img && sleep 2 && docker exec oqs-httpd bash -c 'echo  "GET /" | /opt/openssl/bin/openssl s_client -CAfile CA.crt -curves frodo640aes -crlf -connect localhost:4433')
+    - run: 
+        name: nginx
+        command: .circleci/git_no_checkin_in_last_day.sh || (cd nginx && docker build -t oqs-nginx-img . && docker run --detach --rm --name oqs-nginx oqs-nginx-img && sleep 2 && docker exec oqs-nginx /opt/openssl/apps/openssl s_client -CAfile CA.crt -curves kyber512 -connect localhost:4433)
+    - run:
+        name: Push images
+        command: docker tag oqs-curl $DOCKER_LOGIN/curl-generic && docker push $DOCKER_LOGIN/curl-generic && docker tag oqs-httpd-img $DOCKER_LOGIN/httpd-generic && docker push $DOCKER_LOGIN/httpd-generic && docker tag oqs-nginx-img $DOCKER_LOGIN/nginx-generic && docker push $DOCKER_LOGIN/nginx-generic
+
+
+# As of now (Mar 2020) CircleCI does not support building current OSX docker images... :-(
```
xvzcf: Could you remove this commented-out text and open an issue to keep track?

.circleci/git_no_checkin_in_last_day.sh
```diff
@@ -0,0 +1,39 @@
+#!/bin/bash
+
```
xvzcf: Could you add `set -e` here?

.circleci/git_no_checkin_in_last_day.sh
```diff
@@ -0,0 +1,39 @@
...
+# Detects whether there has been a Git commit in the last day on this
+# branch. Returns 1 if there has been a commit, 0 if there has not.
+
+r=`git log --name-only --since="1 day ago" -n 2`
```
xvzcf: Could you replace backticks for shell command invocations with `$(...)`?

.circleci/config.yml
```diff
@@ -0,0 +1,98 @@
...
+    environment:
+      IMAGE: openquantumsafe/ci-ubuntu-bionic-x86_64:latest
+#  macOS:
+#    <<: *macjob
+
+workflows:
+  version: 2
+  build:
+    jobs:
+      - debian-buster-amd64:
```
xvzcf: Should we make both jobs nightly ones?

.circleci/git_no_checkin_in_last_day.sh
baentsch: Done

.circleci/git_no_checkin_in_last_day.sh
baentsch: Done

.circleci/config.yml
baentsch: Done: #13

.circleci/config.yml
baentsch: Hmm - I'd think it'd be sensible to keep one running directly at checkin to get direct feedback if all is fine. But I switched order to also validate the Ubuntu job now as we know the Debian job works.

xvzcf: True. We should modify the Push images step so that we don't push new images every time a commit is pushed to a PR, perhaps using an environment variable. We can have both the jobs be nightlys that actually push images to the Docker Hub.

baentsch: Well, each job over-writes the previous' docker image if no errors occurred, so I'd argue only one should run nightly, i.e., leave as-is. Without comments to the opposite, I'll merge (say after today's call).

PR #31, opened 2020-05-26
httpd switched to alpine

baentsch:
Another 110MB saved.

Test run at https://circleci.com/gh/baentsch/oqs-software/tree/mb-httpd-alpine

### DIFF
--- a/httpd/Dockerfile
+++ b/httpd/Dockerfile
@@ -19,7 +19,7 @@ ARG HTTPD_VERSION=2.4.43
 ARG MAKE_DEFINES="-j 2"
 
 
-FROM debian:buster-slim as intermediate
+FROM alpine as intermediate
 # Take in global args
 ARG LIBOQS_BUILD_DEFINES
 ARG OPENSSL_PATH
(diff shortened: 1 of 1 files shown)


httpd/Dockerfile
```diff
@@ -92,17 +86,23 @@ RUN set -x && \
...
+#
+RUN apk add pcre-dev expat-dev
+#
 # Only retain the ${*_PATH} contents in the final image
 COPY --from=intermediate ${HTTPD_PATH} ${HTTPD_PATH}
+# copy over manually build libapr{util}
+COPY --from=intermediate /usr/local/apr/lib  /usr/local/apr/lib
+
+# Some size optimization:
+RUN rm -rf ${HTTPD_PATH}/bin/ab
```
dstebila, line 103: I'm not a Docker expert, but won't the effect of this be that the filesystem layer prior to the `RUN rm` command will still have ab in it?  Instead shouldn't you delete ab back in the intermediate image so it's never copied over?

baentsch: Yes, that's even better. Thanks!

PR #32, opened 2020-06-02
Updated Chromium instructions.

xvzcf:
This can go in once https://github.com/open-quantum-safe/boringssl/pull/22 does.

### DIFF
--- a/chromium/README.md
+++ b/chromium/README.md
@@ -1,12 +1,10 @@
-This directory contains instructions and corresponding patches to build the Chromium web browser using the [OQS BoringSSL fork](https://github.com/open-quantum-safe/boringssl), thereby enabling Chromium to use the quantum-safe key exchange algorithms provided by liboqs.
-
-The fork, and by extension Chromium, at present only support the use of quantum-safe key exchange algorithms in TLS 1.3. Furthermore, the instructions have only been tested on an Ubuntu 19.04 x86_64 machine.
+This directory contains instructions and corresponding patches to build the Chromium web browser using the [OQS-BoringSSL fork](https://github.com/open-quantum-safe/boringssl), thereby enabling Chromium to use quantum-safe key exchange algorithms. Note that these instructions have only been tested on an Ubuntu 19.10 x86_64 machine and apply at present only to a select subset of quantum-safe key-exchanges.
 
 0. Ensure the system requirements listed [here](https://chromium.googlesource.com/chromium/src/+/master/docs/linux/build_instructions.md#System-requirements) are met.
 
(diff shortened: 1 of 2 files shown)


chromium/README.md
```diff
@@ -17,16 +15,17 @@ The fork, and by extension Chromium, at present only support the use of quantum-
...
 
-To verify that Chromium can perform a TLS 1.3 handshake using a post-quantum key exchange:
+To verify that Chromium can perform a TLS 1.3 handshake using a post-quantum key exchange and authentication algorithm:
 
+0. Navigate to `<CHROMIUM_ROOT>`, and start Chromium by executing `./out/Default/chrome --ignore-certificate-errors`
 1. Navigate again to the `<CHROMIUM_ROOT>/third_party/boringssl/src` folder, and build OQS-BoringSSL as a standalone project by following the instructions in that directory's `README.md` file.
-2. Then, in the `build` directory, run `./tool/bssl server -curves oqs_<KEX> -accept localhost:4433`, where `<KEX>` is any key-exchange algorithm listed [here](https://github.com/open-quantum-safe/boringssl#supported-algorithms).
-3. In the Chromium browser, load `https://localhost:4433`.
+2. Then, in the `build` directory, run `./tool/bssl server -accept 4433 -www -loop -curves <KEX>`, where `<KEX>` can be any key-exchange algorithm that are both listed [here](https://github.com/open-quantum-safe/boringssl#supported-algorithms) and are listed in the array [here](https://github.com/open-quantum-safe/boringssl/blob/master/ssl/t1_lib.cc#L375).
```
baentsch, line 30: This is pretty complex (to understand): Why check 2 locations? Am I right assuming this is again the 'manual default KEX reduction'? Wouldn't it be reasonable to _only_ look at https://github.com/open-quantum-safe/boringssl/blob/master/ssl/t1_lib.cc#L375 (it's a subset of all algorithms, right)? Also, would it be sensible to a facility such as https://github.com/open-quantum-safe/openssl#default-algorithms-announced ?

xvzcf: The first location provides the names of the algorithms to pass to the `bssl server` command, and the second location lists all the algorithms that can be specified. 

I've slightly amended this sentence to make it more clear, and added a note linking to [this page](https://github.com/open-quantum-safe/boringssl/wiki/Implementation-Notes) (to be filled out) that explains why the algorithms supported by "default" are a subset of all supported algorithms.

I don't know if it's worth having an `OQS_DEFAULT_GROUPS` option for BoringSSL at the moment, as I don't think it would make building Chromium that much friendlier (the option would have to be specified somewhere around [here](https://chromium.googlesource.com/chromium/src.git/+/refs/heads/master/third_party/boringssl/BUILD.gn#38), and I haven't verified that that is all that is needed). The option might be worth having if we do start using BoringSSL for cURL.

Issue #43, opened 2020-07-20
Chromium fails to connect to QSC-signature protected server


baentsch:
"Weak key" error when connecting Chromium to QSC-sig cert protected test port, e.g., https://test.openquantumsafe.org:6471/

baentsch:
Will close when PR #51 gets merged.

PR #45, opened 2020-07-20
enhanced patch to get QSC certs accepted

baentsch:
Addresses Issue #43 . @xvzcf Please check whether I did the patch correctly/in the same way you (wanted) it created. Hence only Draft PR.

### DIFF
--- a/chromium/qsc-certs.patch
+++ b/chromium/qsc-certs.patch
@@ -1,3 +1,28 @@
+From 021cbbf2e551386c236fb3e9fa97f95be894d278 Mon Sep 17 00:00:00 2001
+From: "Ubuntu 19.10" <57787676+baentsch@users.noreply.github.com>
+Date: Mon, 20 Jul 2020 10:23:43 +0200
+Subject: [PATCH] enable QSC certs
+
+---
+ net/cert/internal/simple_path_builder_delegate.cc | 4 ++--
+ third_party/boringssl/BUILD.gn                    | 5 +++--
+ 2 files changed, 5 insertions(+), 4 deletions(-)
+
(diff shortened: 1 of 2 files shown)


xvzcf:
Looks fine to me.

chromium/qsc-certs.patch
```diff
@@ -1,3 +1,28 @@
...
+--- a/net/cert/internal/simple_path_builder_delegate.cc
++++ b/net/cert/internal/simple_path_builder_delegate.cc
+@@ -110,8 +110,8 @@ bool SimplePathBuilderDelegate::IsPublicKeyAcceptable(EVP_PKEY* public_key,
+     return true;
+   }
+ 
+-  // Unexpected key type.
+-  return false;
++  // Accept all keys for QSC;
++  return true;
```
dstebila, line 22: Accepting all keys seems very permissive. Is there a risk here that too much is being accepted?  Can we add a return true for the OQS algorithms but still have a return false for the remainder?

baentsch: I also thought about that -- but then had to accept I don't understand BoringSSL already good enough to facilitate such check. @xvzcf: Do you have a suggestion?

dstebila:
@xvzcf Can this be updated now that open-quantum-safe/boringssl#28 has landed?

baentsch:
@xvzcf Do you have an update to this PR with your BoringSSL-algorithm check or should I try to build one (rebasing everything)? Do we rebase to 0.4.0 or latest?

xvzcf:
I do have an update, I'll put it up tomorrow or the day after and we can take it from there.

baentsch:
Replaced by #51

PR #51, opened 2020-09-03
enable QScerts in Chromium

baentsch:
Logic understood: Looks good to me -- certainly more secure than #45. I'm in the process of building this -- but this may take a day or two on my current machine...

### DIFF
--- a/chromium/oqs-mods.patch
+++ b/chromium/oqs-mods.patch
@@ -1,3 +1,18 @@
+diff --git a/net/cert/internal/simple_path_builder_delegate.cc b/net/cert/internal/simple_path_builder_delegate.cc
+index e4259ee6eb39..74e8158edfda 100644
+--- a/net/cert/internal/simple_path_builder_delegate.cc
++++ b/net/cert/internal/simple_path_builder_delegate.cc
+@@ -35,6 +35,10 @@ bool IsAcceptableCurveForEcdsa(int curve_nid) {
+       return true;
+   }
+ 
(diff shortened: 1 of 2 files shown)


chromium/oqs-mods.patch
```diff
@@ -1,3 +1,18 @@
+diff --git a/net/cert/internal/simple_path_builder_delegate.cc b/net/cert/internal/simple_path_builder_delegate.cc
+index e4259ee6eb39..74e8158edfda 100644
+--- a/net/cert/internal/simple_path_builder_delegate.cc
++++ b/net/cert/internal/simple_path_builder_delegate.cc
+@@ -35,6 +35,10 @@ bool IsAcceptableCurveForEcdsa(int curve_nid) {
+       return true;
+   }
+ 
++  if (IS_OQS_PKEY(curve_nid)) {
++      return true;
```
dstebila, line 10: Just to make sure I understand the context, this is not a fail-open change?

baentsch: Full context: 
```
bool IsAcceptableCurveForEcdsa(int curve_nid) {
  switch (curve_nid) {
    case NID_X9_62_prime256v1:
    case NID_secp384r1:
    case NID_secp521r1:
      return true;
  }

  if (IS_OQS_PKEY(curve_nid)) {
      return true;
  }

  return false;
}
```
--> Thus, only classic EC curves and OQS "curves" allow this to succeed.

dstebila (approved):

PR #58, opened 2020-10-19
Refactored CI.

xvzcf:


.circleci/config.yml
```diff
@@ -10,90 +15,95 @@ localCheckout: &localCheckout
...
-#  macOS:
-#    <<: *macjob
+  linux_x64_openssl:
+    description: A template for building and pushing OQS demo Docker VMs
+    parameters:
+      CONTAINER:
+        description: "The docker container to use."
+        type: string
+    docker:
+      - image: << params.CONTAINER >>
```
baentsch, line 26: should this read "parameters"?

baentsch (changes requested):
Conceptually I get it. But what looks like a typo seems to fail CI...

baentsch (approved):
Using only ubuntu to create demo images is surely good enough. I shall be curious to see the images pushed when this is on master.

PR #59, opened 2020-10-26
main branch renaming

baentsch:
First step to eliminate master branch. If OK, please also set as default.


dstebila:
Oops, should I have merged this before changing default branch to main?

baentsch:
Main thing it's in main (Sorry, just couldn't resist :-) All looks good. Thanks for setting to default.

PR #65, opened 2020-12-15
Added OQS-OpenSSH demo

xykro:
Hi @baentsch and @dstebila

Referring to issue #54. 

As we talked about, we have (the [ISE institute](https://www.fhnw.ch/en/about-fhnw/schools/school-of-engineering/institutes/institute-for-sensors-and-electronics) from [FHNW](https://www.fhnw.ch/en)) have created a docker demo image that uses your OQS-OpenSSH implementation.

The [README.md](https://github.com/fhnw-ise-qcrypt/oqs-demos/blob/pull-request/openssh/README.md) contains setup and general information while the [USAGE.md](https://github.com/fhnw-ise-qcrypt/oqs-demos/blob/pull-request/openssh/USAGE.md) describes some more advanced use cases, like using this demo to connect to a remote OQS-OpenSSH instance.

I've added a new job for [CircleCI](https://github.com/fhnw-ise-qcrypt/oqs-demos/blob/pull-request/.circleci/config.yml) for this demo image. The test it does is connecting to itself (localhost) using chosen PQC algorithms via OQS-SSH. Defaults are p256-dilithium2 and ecdh-nistp384-kyber-1024, chosen more or less randomly. Those algorithms may be changed by passing the SIG and/or KEM variables to the docker run command.
Return code is 0 if the connection to localhost using the ubkey method was successful, 1 otherwise.

I have **not** added OQS-OpenSSH to the "Push all images" step, I reckoned as you will be hosting the image I'll let you set that up properly.

Git history is squashed down to 1 commit, I do not want to bring my mess over :-)

Let me know if any adjustments are required.


openssh/USAGE.md
```diff
@@ -0,0 +1,128 @@
...
+
+## Slightly more advanced usage options
+
+### Change install location
+
+The OQS-OpenSSH binaries, configuration files and host keys are located in the default install location `/opt/oqs-ssh/`. This location can be changed at build time with `--build-arg INSTALL_DIR=</path/to/new/location>`.
+
+### Access the man pages of oqs-ssh
+
+Man pages for oqs-ssh are installed in `<INSTALL_DIR>/share/man` and can be viewed by for example `man -l <INSTALL_DIR>/share/man/man1/ssh.1` or `man -l <INSTALL_DIR>/share/man/man5/ssh.5`. Note that those man pages are **not** different from the original ssh man pages. So it might be easier to consult them on your local system or the internet.
```
baentsch, line 24: suggest to change these sample commands to contain the full "docker-wrapper" command line so it can be executed immediately (cut-copy-paste)

openssh/USAGE.md
```diff
@@ -0,0 +1,128 @@
...
+
+       docker run -dit --network host --name oqs-ssh --restart unless-stopped oqs-openssh-img
+
+The `--network host` option will attach the container directly to your host's network, sharing its IP. The sshd in the container is now accessible from the outside using the host's IP address and the specified port (2222 per default).
+
+Be aware that the port 2222 also needs to be open in any firewall's there may be!
+
+### Set up the client
+
+For the client side, you need to compile yourself a docker image as described in the 'Quick start' section of the [README.md](README.md). You then run it as normal with
```
baentsch, line 100: Why can't one use the (pre-existing/documented) docker image for this? The goal of this document is to be standing alone at hub.docker.com, just like e.g., https://hub.docker.com/r/openquantumsafe/nginx

openssh/key-regen.sh
```diff
@@ -0,0 +1,78 @@
...
+    fi
+done
+# echo " done!"
+
+# Regenerate existing host keys
+HOST_KEY_DIR=$(echo $OQS_INSTALL_DIR | sed 's:/*$::')
+echo -e "\nGenerating host key files as configured in $OQS_INSTALL_DIR/sshd_config (all HostKeyAlgorithms):"
+
+# Get algorithms from sshd_config
+IFS=',' read -ra HOST_KEY_ALGS <<< $(sed -n "s/^hostkeyalgorithms[ \t=]*//Ip" ${OQS_INSTALL_DIR}/sshd_config)
```
baentsch, line 48: Do I understand this correct in that one has to manually update the set of supported algorithms in sshd_config as and when the algorithm list supported upstream changes? Might it be worth while thinking about some automation for this? But then again, it's not a regular occurrence...

openssh/README.md
```diff
@@ -0,0 +1,106 @@
...
+
+[Install Docker](https://docs.docker.com/install) and run the following commands in this directory:
+
+1. Run `docker build -t oqs-openssh-img .` This will generate the image with the name `oqs-openssh-img`
+2. `docker run --name oqs-openssh-server -ditp 2222:2222 --rm oqs-openssh-img`
+This will start a docker container that has sshd listening for SSH connections on port 2222 and this port is forwarded to and accessible via `localhost:2222`.
+3. `docker run --rm --name oqs-openssh-client -dit oqs-openssh-img` will start a docker container with the same properties as `oqs-openssh-server` except the port 2222 is not published.
+4. You can hop on either of those two containers as a non-root user (oqs) to use the built-in OQS-OpenSSH binaries or do other shenanigans by typing
+`docker exec -ti -u oqs oqs-openssh-server /bin/sh`
+Of course adjust the container's name accordingly if hopping onto the client (==> `oqs-openssh-client`).
```
baentsch, line 15: again, suggest using a "cut-copy-paste" ready line instead of just 'oqs-openssh-client'

openssh/README.md
```diff
@@ -0,0 +1,106 @@
...
+This directory contains a Dockerfile that builds the [OQS OpenSSH fork](https://github.com/open-quantum-safe/openssh), which allows to establish a quantum-safe SSH connection using quantum-safe keys and quantum-safe authentication.
+
+## Quick start
+
+[Install Docker](https://docs.docker.com/install) and run the following commands in this directory:
+
+1. Run `docker build -t oqs-openssh-img .` This will generate the image with the name `oqs-openssh-img`
+2. `docker run --name oqs-openssh-server -ditp 2222:2222 --rm oqs-openssh-img`
+This will start a docker container that has sshd listening for SSH connections on port 2222 and this port is forwarded to and accessible via `localhost:2222`.
+3. `docker run --rm --name oqs-openssh-client -dit oqs-openssh-img` will start a docker container with the same properties as `oqs-openssh-server` except the port 2222 is not published.
```
baentsch, line 12: Would it be better to leave away the numbers 3 and 4 (and rephrase somewhat), as these are an alternative (3) and an optional subsequent step (4), but not essentially required to run through the demo?

openssh/README.md
```diff
@@ -0,0 +1,106 @@
...
+
+The Dockerfile 
+- obtains all source code required for building the quantum-safe cryptography (QSC) algorithms and the [QSC-enabled version of OpenSSH (7.9-2020-08_p1)](https://github.com/open-quantum-safe/openssh/releases/tag/OQS-OpenSSH-snapshot-2020-08)
+- builds all libraries and applications
+- by default starts the openssh daemon\*
+- creates a second user `oqs` with the default password `oqs.pw`
+- by default creates host keys based on the enabled host key algorithms in `sshd_config`\*
+- by default creates identity keys based on the config file `ssh_config`\*
+
+\*those steps are executed when the image is startet.
```
baentsch, line 44: typo: "startet" -> "started"

baentsch:
@xykro Thanks very much for this work (incl. the squashing -- makes it very readable) and the nice use case in USAGE.md! I also like very much the explanations and "helpful hints" if something goes wrong. I provided some suggestions for improvements above. Most notably, I'd be grateful for considering updates to the USAGE.md such that that can be cut-copy-pasted straight to DockerHub when everything builds OK within the OQS CCI chain (from where it shall be auto-pushed if all tests pass: You could actually retain the "auto-push" logic if you'd set the environment variable TARGETNAME within your CCI environment to a docker hub prefix you might control, e.g., "fhnw-ise-qcrypt" to have the exact same setup).
For a quick glance-over regarding the openssh-specifics would @xvzcf or @christianpaquin want to take a deeper look?

xykro:
@baentsch Many thanks for your feedback! I will address the points you mentioned.

CCI shows above that all tests passed, but if you look closely all BUT OpenSSH was tested. I registered a new job *ubuntu_x64_openssh* because I did not want to mess with your other tests you mention to be "some kind of SSL-using", which SSH is not. Why CCI did not automatically run the other job I don't know, on my side (CCI) it just did it. But I guess for simplicitiy's sake it is easier to add OpenSSH to the other tests and omit the fact that it is not SSL based. So it would be retested upon a change of OpenSSL with the others, which would not be the end  of the world.

Long story short: Should I integrate the test job into your existing job or keep the new one I created?

openssh/USAGE.md
xykro: I was not aware of the full purpose of the USAGE.md, that it will be used directly on hub.docker.com. I will adjust all mentions of the image accordingly.

openssh/key-regen.sh
xykro: You do understand this correctly.

The reason I ended up with this version is that I didn't automatically want to generate all host keys with `ssh-keygen -A`  (which would not solve the problem with the config files) because it takes quite some time especially with the rainbow keys. For the same reason I use the build target `make install-nokeys` when compiling ssh while building the image, it speeds up building considerably. So I needed some place to configure the algorithms that are used per default, because this affects the generation of the keys as well as the configuration of sshd and ssh.

One possibility would be adding a feature to specify the algorithms at build time and automatically update sshd_config and ssh_config accordingly in the Dockerfile, using sed or something similar. Thinking about it, I will have a short try with this.

And I just realized that I diverged from what you were actually asking: For now I don't see another possibility to automate this without more complex scripting that checks upstream algorithms changes and applies them. But as you say, it's quite irregular and as we have already progressed quite far in the selection of PQC algorithms, there probably won't be that many changes before the new standards are released.

baentsch:
> Should I integrate the test job into your existing job or keep the new one I created?

OK you have a point. No, I'd then prefer to leave it separate as you suggest. It would be good to augment it with a similar push step as in the openssl job, though.

openssh/key-regen.sh
baentsch: Agreed. But maybe a footnote stating this would be good. Or maybe even better a test that would/will break if the upstream algorithm list changes.

christianpaquin: If you are aiming to only update this for major releases of oqs/openssh-oqs (e.g., v0.4.0 in this case), than the manual process might be ok. Agreed with @baentsch that a failsafe test would be good.

openssh/README.md
```diff
@@ -0,0 +1,106 @@
...
+
+**Note for the interested**: The build process is two-stage with the final image only retaining all executables, libraries and include-files to utilize OQS-enabled openssh.
+
+The re-generation of the host and identity keys happens via the script [key-regen.sh](key-regen.sh) that is called with a `CMD` command at the end of the [Dockerfile](Dockerfile). The script checks if the required key already exist and gerenates it if necessary.
```
christianpaquin, line 48: "The script checks if the required key already exist and gerenates it if necessary."

What happens if it already exists? The name of the script makes it sound like it will be overwritten. If it's not the desired outcome, maybe the script should be named `key-gen.sh`. If it is as expected, I'd remove the "if already exist" mention in the text.

openssh/README.md
```diff
@@ -0,0 +1,106 @@
...
+    --prefix=${INSTALL_DIR} \
+    --sysconfdir=${INSTALL_DIR} \
+    --with-liboqs-dir=/opt/ossh-src/oqs \
+    --with-mantype=man \
+    ${OPENSSH_BUILD_OPTIONS}
+```
+These parameters will be overridden if specified again in `OPENSSH_BUILD_OPTIONS`.
+
+`/opt/ossh-src/oqs` is the location the intermediate build in the Dockerfile writes the compiled liboqs binaries to. This should not be changed, as it does not influence the final docker image.
+### MAKE_DEFINES
```
christianpaquin, line 94: Cosmetic: Add a newline before ### MAKE_DEFINES

openssh/README.md
```diff
@@ -0,0 +1,106 @@
...
+
+Allow setting parameters to `make` operation, e.g., `-j nnn` where nnn defines the number of jobs run in parallel during build. 
+
+The default is conservative and known not to overload normal machines (default: `-j 2`). If one has a very powerful (many cores, >64GB RAM) machine, passing larger numbers (or only `-j` for maximum parallelism) speeds up the build considerably.
+
+### OQS_USER
```
christianpaquin, line 100: Any plans of supporting key-based client auth?

openssh/USAGE.md
```diff
@@ -0,0 +1,128 @@
...
+
+The OQS-OpenSSH binaries, configuration files and host keys are located in the default install location `/opt/oqs-ssh/`. This location can be changed at build time with `--build-arg INSTALL_DIR=</path/to/new/location>`.
+
+### Access the man pages of oqs-ssh
+
+Man pages for oqs-ssh are installed in `<INSTALL_DIR>/share/man` and can be viewed by for example `man -l <INSTALL_DIR>/share/man/man1/ssh.1` or `man -l <INSTALL_DIR>/share/man/man5/ssh.5`. Note that those man pages are **not** different from the original ssh man pages. So it might be easier to consult them on your local system or the internet. 
+
+Direct links to man pages: [ssh(1)](https://linux.die.net/man/1/ssh), [sshd(8)](https://linux.die.net/man/8/sshd), [ssh_config(5)](https://linux.die.net/man/5/ssh_config), [sshd_config(5)](https://linux.die.net/man/5/sshd_config)
+
+## Seriously more advanced usage options
```
christianpaquin, line 28: I would drop the value-judgement and just call this section "Advanced usage options" :)

openssh/USAGE.md
```diff
@@ -0,0 +1,128 @@
...
+
+### Choosing the algorithms
+
+For a list of all algorithms see [here](https://github.com/open-quantum-safe/openssh#supported-algorithms). It is recommended to only use hybrid algorithms to maintain established classical security. The post-quantum safe algorithms have not yet received enough confidence to be relied on as the only security mechanism.
+
+The image's default key-exchange algorithm is `ecdh-nistp384-kyber-1024-sha384@openquantumsafe.org` for key-exchange with `curve25519-sha256` for backwards compatibility. For host and identity key (authentication) algorithms `ssh-p256-dilithium2` and classical `ssh-ed25519` are used. Those algorithms may be changed by adjusting the files `ssh_config` and `sshd_config` respectively. In the built image, those files are located at the default install location (see above). After changing something in `sshd_config`, the sshd must be restarted using `rc-service oqs-sshd restart`. Alternatively, the configuration can be changed **pre**-build by changing [ssh_config](ssh_config) or [sshd_config](sshd_config) and rebuilding the image.
```
christianpaquin, line 34: I don't understand  "backward compat" means in this sentence: "The image's default key-exchange algorithm is `ecdh-nistp384-kyber-1024-sha384@openquantumsafe.org` for key-exchange with `curve25519-sha256` for backwards compatibility." The former alg is already an hybrid; what fallback does x25519 provides?

christianpaquin, line 34: Same question for the ed25519 auth fallback.

openssh/USAGE.md
```diff
@@ -0,0 +1,128 @@
...
+
+The location where `key-regen.sh` is looking for `ssh_config`/`sshd_config` is the install directory of `oqs-ssh`. The [Dockerfile](Dockerfile) puts this location into the variable `OQS_INSTALL_DIR` where it will be accessible from the script.
+## Using oqs-ssh for quantum-safe remote access with minimal intrusion
+
+One use case of quantum-safe ssh running in docker could be accessing a remote system without messing with its ssh(d) installation or other parts of the system you maybe don't want to interact with. This means minimal intrusion and everything can easily be removed again. This is done by running this docker image on said system and sharing its network space. Thus it is possible to access host ports from **within** the docker container. Normally, the use case of docker is one of isolation with some shared directories and published ports at max. So this solution works around the usual docker limitations.
+
+Additionally, it is advised to **change the default username and password** when building the image because your plan is to expose it to the world.
```
christianpaquin, line 58: Is there a way to force a deployer to explicitly pick a new password rather than suggesting it?

christianpaquin (changes requested):
Thanks for submitting this! I have some questions about and suggestions on the documentation.

christianpaquin:
I tried building the containers and connecting from the client to server; it worked without a glitch!

openssh/key-regen.sh
xykro: I added the option to change the install target at build time, and parametrized the liboqs and oqs-openssh version in the Dockerfile, so it is clearer what version is being used.

> Agreed. But maybe a footnote stating this would be good. Or maybe even better a test that would/will break if the upstream algorithm list changes.

I will mention this in the Readme, and also in form of a comment in the Dockerfile.  As I explicitely check out the 0.4.0 release, nothing would change without a user actually making a change AFAIK. So I suppose mentioning it should suffice. If you agree I will leave it at this, also because I don't have an angle on how to automatically test this, would this be a part of the CCI config? It could notify that there is a new release of liboqs, so it wouldn't slip through.

openssh/Dockerfile
```diff
@@ -0,0 +1,133 @@
...
+WORKDIR /opt
+RUN git clone --depth 1 --branch 0.4.0 https://github.com/open-quantum-safe/liboqs && \
+    git clone --depth 1 --branch OQS-OpenSSH-snapshot-2020-08 https://github.com/open-quantum-safe/openssh ossh-src;
+
+# build liboqs static (does not work with shared lib!)
+WORKDIR /opt/liboqs
+# RUN mkdir build && cd build && cmake .. ${LIBOQS_BUILD_DEFINES} -DBUILD_SHARED_LIBS=ON -DCMAKE_INSTALL_PREFIX=/opt/ossh-src/oqs && make ${MAKE_DEFINES} && make install
+RUN mkdir build-static && cd build-static && cmake .. ${LIBOQS_BUILD_DEFINES} -DCMAKE_BUILD_TYPE=${LIBOQS_BUILD_TYPE} -DBUILD_SHARED_LIBS=OFF -DCMAKE_INSTALL_PREFIX=/opt/ossh-src/oqs && make ${MAKE_DEFINES} && make install
+
+# build OQS-OpenSSH, this will automatically generate host keys for all SIG algorithms
```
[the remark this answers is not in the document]
xykro: Oupsi, that's an older comment that slipped through :) Thanks for the careful reading.

openssh/README.md
xykro: It actually is supported, like with standard ssh you may add your public identity key (that is generated in `key-regen.sh` based on `ssh_config`)  to the server's `known_keys` file.

Might be worth mentioning in the docs though.

openssh/USAGE.md
xykro: The fallback to standard ssh. But thinking about it, as this *is* supposed to be a quantum-safe demo of ssh, it does not make much sense to support that. You can always fall back on the standard ssh installation if available.

xykro:
> I tried building the containers and connecting from the client to server; it worked without a glitch!

Happy to hear that! Thanks for trying it out!

openssh/USAGE.md
xykro: Same answer as for the x25519 fallback 😉

openssh/USAGE.md
xykro: I am not aware of a way to do this, and a really quick googlin' didn't bring any revelation. 

This actually brings up another topic I must admit I am not really happy with. As I have discussed with @baentsch some time ago, it may make more sense to switch to the oqs user in the Dockerfile and never start the container as root. But this makes then running the `sshd` a little more complicated. There are a few topics in the wild internet talking about running `sshd` without root privileges, but some of those solutions are deprecated and it does not seem common. I have tried some things, but nothing really worked to my satisfaction. And also from a security standpoint I it seems it is even recommended practice to run the `sshd` as root. So I sticked to using root. That basically means the oqs user is only used for remote login and not for local container management, which I can live with.

openssh/USAGE.md
xykro: I blatantly stole that from @baentsch's [nginx USAGE.md](https://github.com/open-quantum-safe/oqs-demos/blob/main/nginx/USAGE.md#seriously-more-advanced-usage-options). I liked the humor and stuck with it 😃

openssh/README.md
xykro: True that. No it does not regenerate keys, or not anymore. Before I started using the `install-nokeys` build target it was regenerating keys because a Docker image with hard coded keys ain't really best practice from a security standpoint.

Said that I will rename it to reflect what it is actually doing.

openssh/README.md
```diff
@@ -0,0 +1,106 @@
...
+usermod -aG docker <user>
+newgrp docker
+```
+The first command adds user `<user>` to the group `docker`, and the second simulates a logout/login, so you don't have to do a re-login.
+
+### Connect from container to container via OQS-SSH
+
+1. To connect to another docker container, we first need to create a docker network by typing `docker network create oqs-openssh-net`.
+
+1. Then connect the containers to this network by typing `docker network connect oqs-openssh-net <name-of-container>` where `<name-of-container>` is once the server's and once the client's name. Obviously this does not work if the containers are not yet running, in this case refer to the [Quick Start](README.md#quick-start) section.
```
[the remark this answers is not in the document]
xykro: I will rewrite this section to make it more clear, the wording is quite confusing.

README.md
```diff
@@ -10,11 +10,12 @@ In most cases, Dockerfiles encode the instructions for ease-of-use: Just do `doc
...
 | **Apache httpd** | [Github: oqs-demos/httpd](httpd)       | [Dockerhub: openquantumsafe/httpd](https://hub.docker.com/repository/docker/openquantumsafe/httpd)                           |
 | **nginx**        | [Github: oqs-demos/nginx](nginx)       | [Dockerhub: openquantumsafe/nginx](https://hub.docker.com/repository/docker/openquantumsafe/nginx)                           |
 | **Chromium**     | [Github: oqs-demos/chromium](chromium) | [Binary for Ubuntu 18.04](https://github.com/open-quantum-safe/oqs-demos/releases/download/v0.4.0/chromium-ubuntu-0.4.0.tgz) |
-| **HAproxy**        | [Github: oqs-demos/haproxy](haproxy)       | [Dockerhub: openquantumsafe/haproxy](https://hub.docker.com/repository/docker/openquantumsafe/haproxy)                           |
+| **HAproxy**      | [Github: oqs-demos/haproxy](haproxy)   | [Dockerhub: openquantumsafe/haproxy](https://hub.docker.com/repository/docker/openquantumsafe/haproxy)                       |
+| **OpenSSH**      | [Github: oqs-demos/openssh](openssh)   | *Dockerhub: openquantumsafe/openssh WIP*                                                                                     |
```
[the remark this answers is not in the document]
xykro: This is because there's not yet a push step in the CCI config for this image. I will add this step referring to @baentsch above and remove the WIP tag and make it an actual link.

openssh/USAGE.md
christianpaquin: I didn't realize @baentsch was using it :) All good then, if we have precedent!

openssh/USAGE.md
christianpaquin: I see (and I understood later after reading more of the documentation). Fallbacks might be problematic in security (e.g., downgrade attacks), and since this is indeed a PQ project, one would assume there is a need for the PQ algs. One scenario could be a server serving multiple clients, some PQ, some not; but perhaps we should explain how to add a classical-only scheme vs. making it a default.

openssh/USAGE.md
christianpaquin: I'm not a docker expert so I'm not sure what's the best course of action here. Following recommended practice is good, so I'm happy with it if we think it's the best ( slash not-overly-complicated ) option.

openssh/USAGE.md
xykro: Yes indeed. I will adapt this.

xykro:
I handled most of you request/applied most of the things I'd do. Everything else I will visit beginning next week.

christianpaquin:
> I handled most of you request/applied most of the things I'd do. Everything else I will visit beginning next week.

Ok, tag me and @baentsch when you are done, we'll take a final look.

xykro:
@christianpaquin and @baentsch 

With some delay I applied some last improvements and did some final testing.

Let me know your thoughts and I hope we will be able to complete this pull-request soon.

christianpaquin:
I'll take a look next week.

christianpaquin (changes requested):
Documentation changes look good. I would simply ask to add in the readme a full example on how to connect to the ssh client to the server, with explicitly duplicated lines for server and client setup. I followed the instructions (from the [quick start]( docker exec -ti oqs-openssh-client ssh oqs@oqs-openssh-server) and [connect](https://github.com/fhnw-ise-qcrypt/oqs-demos/blob/pull-request/openssh/README.md#connect-from-container-to-container-via-oqs-ssh) readme sections) trying to create the client and server images, but the client exits after creating keys. I got it to work last time, but not this time.

I'll also let @baentsch take a final look since he is our docker expert.

baentsch:
Things look seriously improved, incl. covering lots of possible problems. Very nice.

> I still had the purpose of the USAGE.md and README.md files confused

I tend to agree: `README.md` still contains quite a bit of usage information, e.g., the whole "### Connect from container to container via OQS-SSH" section: In my eyes, a person only doing `docker run` (say, off docker hub) should understand anything s/he can do by reading `USAGE.md` (from quick start to advanced usage). `README.md` should (only) explain all build options (and point to `USAGE.md` as/where appropriate).

From the contents perspective, I'd kind of prefer a stronger algorithm than dilithium2-hybrid as default.

Nit-picks: Some files git complains about as not containing a final newline.

>  @baentsch take a final look

Will do that once the docs are truly final. My test will be going through everything both as a user & builder and expect all sample steps to work (possibly trying some alternatives, too :-). @xykro please tag me when this is done:

> Will clean that up and follow up with one last commit

xykro:
Thank you guys for your time reviewing my work. And forgive me in case I'm using the Github-PR-Tool in unusual ways, it's my first time doing a PR and sometimes I am a bit confused between commenting/responding/reviewing and so on. So many features! 😄 

---
> I followed the instructions (from the [quick start]( docker exec -ti oqs-openssh-client ssh oqs@oqs-openssh-server) and [connect](https://github.com/fhnw-ise-qcrypt/oqs-demos/blob/pull-request/openssh/README.md#connect-from-container-to-container-via-oqs-ssh) readme sections) trying to create the client and server images, but the client exits after creating keys. I got it to work last time, but not this time.

@christianpaquin I agree, there was something messed up. Clarified this one.

---
> I tend to agree: `README.md` still contains quite a bit of usage information, e.g., the whole "### Connect from container to container via OQS-SSH" section: In my eyes, a person only doing `docker run` (say, off docker hub) should understand anything s/he can do by reading `USAGE.md` (from quick start to advanced usage). `README.md` should (only) explain all build options (and point to `USAGE.md` as/where appropriate).

@baentsch I think now I am at a point where I can clearly see each files purpose. Thus far it was kind of foggy in my perception. Will address this and planning on commiting later today.

---
> From the contents perspective, I'd kind of prefer a stronger algorithm than dilithium2-hybrid as default.

@baentsch In the [OQS-OpenSSH-Repository, tag OQS-OpenSSH-snapshot-2020-08](https://github.com/open-quantum-safe/openssh/tree/OQS-OpenSSH-snapshot-2020-08#digital-signature) (which I am using in this docker build), it is stated that, and I quote "only L1 signature and all Rainbow variants are enabled by default". Enabling more algorithms is a more complicated process (which I will mention in my updated README) that requieres adjusting some files after pulling the repo and is so far not supported in the build process. And I don't want to use Rainbows because they are slow as heck and annoying when trying and using the image. Additionally I reckon, that if only those algorithms are enabled by default there may be some reason behind it?

---
> @xykro please tag me when this is done.

@baentsch Will do.

---
One more thing: @baentsch would you quickly review the `.circleci/config.yml` file and check if my configuration for openssh looks like it should in order to work properly? Let me know if anything needs improvement there.

xykro:
@baentsch, @christianpaquin new commit is here!

I changed the structure of README.md and USAGE.md quite substantially, moved a lot of content from README to USAGE and added some additional references back to the README and building the image yourself.

Additionally, I streamlined the key-generation process, so that the requirements for ssh_config and sshd_config are somewhat similar in structure to get the keys one wants. Some new, much more detailed information about the configuration is also available.

And there are some more minor changes I can't remember in detail.

And the best for last: We have got newlines at the end of every file now! 🥳 

Let me know what you think!

baentsch:
> only L1 signature and all Rainbow variants are enabled by default 

Kind of -- only "asterisk'd algs" are (recent change in doc). But you're right: L2's are not enabled by default. My mistake - I was only driven by the wish for "the middle ground" (NIST level 3 :-) Not an (easy) option, you're right -- maybe an issue for a next iteration. So disregard my comment.

> would you quickly review the .circleci/config.yml file and check if my configuration for openssh looks like it should in order to work properly

Sorry I see this only now that you're obviously already running CI --successfully: So I guess I can consider this moot and proceed with final review/trying things out (first thing tomorrow)?

xykro:
> Kind of -- only "asterisk'd algs" are (recent change in doc). But you're right: L2's are not enabled by default. My mistake - I was only driven by the wish for "the middle ground" (NIST level 3 :-) Not an (easy) option, you're right -- maybe an issue for a next iteration. So disregard my comment.

Yes, I saw that they are "asterisk'd" now. Maybe I will automate enabling the other algorithms when benchmarking the ssh connections. Until then, it's L1 algorithms only so far. Even though I'd like to have L3 enabled as well.

> Sorry I see this only now that you're obviously already running CI --successfully: So I guess I can consider this moot and proceed with final review/trying things out (first thing tomorrow)?

I know that CI tests are running and passing, what I don't know is whether pushing the image will work. But I guess we'll see about that. That said: Yes, you may proceed :)

baentsch:
>  I don't know is whether pushing the image will work. 

What I read, it should. Again, you could also test that yourself if you'd register in our own CircleCI namespace the environment variables TARGETNAME, DOCKER_LOGIN and DOCKER_PASSWORD, with contents matching your own docker location, docker hub ID and docker password, respectively: In that case you'd trigger the logic from your account and the resultant image would wind up in your docker hub. But again, this looks good.

Now to the contents: Separation between `README.md` and `USAGE.md` is perfect now and I'll now go through all tests. For that, let me start a review....

openssh/USAGE.md
```diff
@@ -0,0 +1,270 @@
...
+
+The image's default key exchange algorithm is `ecdh-nistp384-kyber-1024-sha384@openquantumsafe.org`. For host and identity key (authentication) algorithms `ssh-p256-dilithium2` is used. Those algorithms may be changed by adjusting the files `ssh_config` and `sshd_config` respectively.
+
+**In `ssh_config` (client side)**
+- `KexAlgorithms`: Comma-separated list of enabled key-exchange algorithms. Priority given by order. Names according to [this KEX naming scheme](https://github.com/open-quantum-safe/openssh#key-exchange).
+- `IdentityKey`: Path to identity key files. One entry for one file, can have multiple entries. SSH will look for those files when connecting to a host. Names of the key files need to be `~/.ssh/id_<SIG>` in order for them to be successfully generated. `<SIG>` is a post-quantum signature algorithm according to [this SIG naming scheme](https://github.com/open-quantum-safe/openssh#digital-signature), with every `-` replace by `_`. A list with of some possible `IdentityKey` values can be found [here](https://github.com/open-quantum-safe/oqs-demos/tree/main/openssh/ssh_config).
```
baentsch, line 182: Nit: "replace" -> "replaced"

openssh/USAGE.md
```diff
@@ -0,0 +1,270 @@
...
+- `HostKeyAlgorithms`: Comma-separated list of the host key types the client accepts from the server. Names according to [this SIG naming scheme](https://github.com/open-quantum-safe/openssh#digital-signature).
+
+**In `sshd_config` (server side)**
+- `KexAlgorithms`: Comma-separated list of enabled key-exchange algorithms. Priority given by order. Names according to [this KEX naming scheme](https://github.com/open-quantum-safe/openssh#key-exchange).
+- `HostKey`: Path to host key files. One entry for one file, can have multiple entries. SSHD will look for those files when offering a host key to a client. Names of the key files need to be `/opt/oqs-ssh/ssh_host_<SIG>_key` in order for them to be successfully generated. `<SIG>` is a post-quantum signature algorithm according to [this SIG naming scheme](https://github.com/open-quantum-safe/openssh#digital-signature), with every `-` replace by `_`. A list with of some possible `HostKey` values can be found [here](https://github.com/open-quantum-safe/oqs-demos/tree/main/openssh/sshd_config).
```
baentsch, line 188: Nit: "replace" -> "replaced"

openssh/USAGE.md
```diff
@@ -0,0 +1,270 @@
...
+- New keys won't have any effect if they are not configured to be used in `HostKeyAlgorithms` or `PubkeyAcceptedKeyTypes` or if other keys have higher priority (first listed in `HostKeyAlgorithms`/`PubkeyAcceptedKeyTypes` has highest priority).
+
+## Automatic key generation
+
+The generation of the host and identity keys happens via the script [key-gen.sh](https://github.com/open-quantum-safe/oqs-demos/tree/main/openssh/key-gen.sh) that is executed automatically every time the container started or run. The script checks if the required key already exist and gerenates it if necessary. This script is called every time the container is started, meaning the first time the container is run (`docker run ...`) and it is startet again (`docker start ...`) after a reboot or a `docker stop ...` command. It checks for existing keys before it generates new ones so it will **never** overwrite an already existing key. It also makes sure `sshd` is startet after at boot time or is restartet after new host keys were generated.
```
baentsch, line 201: Nits: "gerenates" -> "generates"; "startet" -> "started"

openssh/USAGE.md
```diff
@@ -0,0 +1,270 @@
...
+
+Which keys to generate is determined using the configuration files (`ssh_config` and `sshd_config`). A key's necessity is determinded based on the following parameters:
```
baentsch, line 203: Nit: "determinded" -> "determined" or maybe even "The need for a specific key is determined based..."

openssh/USAGE.md
```diff
@@ -0,0 +1,270 @@
...
+
+Enabling more algorithms would require changing [openssh/oqs_templates/generate.yml](https://github.com/open-quantum-safe/openssh/blob/OQS-master/oqs-template/generate.yml) according to [this documentation](https://github.com/open-quantum-safe/openssh/wiki/Using-liboqs-supported-algorithms-in-the-fork#code-generation). Additionally, you need to make sure that the algorithms are enabled in [liboqs](https://github.com/open-quantum-safe/liboqs) as well (see [here for more information](https://github.com/open-quantum-safe/liboqs/wiki/Customizing-liboqs#oqs_enable_kem_algoqs_enable_sig_alg)). Enabling more algorithms in `liboqs` can be done at Docker build time using the build option `LIBOQS_BUILD_DEFINES`. But enabling them in `OpenSSH` would require changing [openssh/oqs_templates/generate.yml](https://github.com/open-quantum-safe/openssh/blob/OQS-master/oqs-template/generate.yml) after checking out `openssh` in the [Dockerfile](https://github.com/open-quantum-safe/oqs-demos/tree/main/openssh/Dockerfile), and this is just not implemented at this moment in time.
```
baentsch, line 226: Suggest adding "in this docker image" at the end of the sentence for further clarification: You did an excellent job describing all that someone needs to know who really wants to do it manually. It also is clear that automating this via docker parameters is significant effort.

openssh/USAGE.md
```diff
@@ -0,0 +1,270 @@
...
+```
+Or you build the image yourself with a different default password from the sources on [Github](https://github.com/openquantumsafe/oqs-demos).
+
+#### Enable classical SSH
+
+Because we wan't to be able to connect to our host that does not run OQS-SSH, we first need to enable classical SSH capabilities for the client **on the host system**.
+
+1. After running the image, run a shell with `docker exec -it oqs-ssh /bin/bash`
+
+2. Run `nano /etc/oqs-ssh/ssh_config` and
```
baentsch, line 126: Now, this unfortunately doesn't work: `/etc/oqs-ssh` doesn't exist. Do you mean `/opt/oqs-ssh`?

openssh/USAGE.md
```diff
@@ -0,0 +1,270 @@
...
+
+1. After running the image, run a shell with `docker exec -it oqs-ssh /bin/bash`
+
+2. Run `nano /etc/oqs-ssh/ssh_config` and 
+   - uncomment the line `# IdentityFile ~/.ssh/id_ed25519`,
+   - add `ssh-ed25519` to `HostkeyAlgorithms` and `PubkeyAcceptedKeyTypes` (comma-separated)
+   - add `curve25519-sha256@libssh.org` to `KexAlgorithms` (comma-separated)
+
+3. Save and exit your editor
+4. Run `rc-service oqs-ssh restart`
```
baentsch, line 132: Fails as follows: 
> rc-service: service `oqs-ssh' does not exist

baentsch (changes requested):
Everything worked perfectly, except for one command and one file location being incorrect. The "merging OQS with classic ssh" piece is great (and works despite the sshd restart failing) after editing the config in the right location (`/opt...`). Please resolve these 2 issues and we're good to merge (the nits I don't care too much about, but you may :-)

xykro:
> Everything worked perfectly, except for one command and one file location being incorrect. The "merging OQS with classic ssh" piece is great (and works despite the sshd restart failing) after editing the config in the right location (/opt...). Please resolve these 2 issues and we're good to merge (the nits I don't care too much about, but you may :-)

@baentsch thanks for your time trying it out, for your feedback and for finding all those nits; happy to resolve them as the devil's in the detail, isn't it 😈 Also fixed the wrong path, should ofc be `/opt/` instead of `/etc/`. And the rc-service `oqs-ssh` should be `oqs-sshd` instead. Addressed all those nits and issues in the last commit.

I think this concludes this in the end rather long back and forth, but I am happy to finally come to a conclusion.

Let me know if goes well with merging!

baentsch (approved):
Thanks very much for your perseverance and the good use case integrating "classic" and OQS sshd, @xykro ! Now LGTM. Will merge tomorrow if @christianpaquin doesn't see issues.

xykro:
> Thanks very much for your perseverance and the good use case integrating "classic" and OQS sshd, @xykro ! Now LGTM. Will merge tomorrow if @christianpaquin doesn't see issues.

I also want to thank you for your perseverance and tireless reviewing and I am happy to have contributed to such a good cause!

PR #66, opened 2021-01-22
fix openssh push command

baentsch:
docker push command fix

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,7 @@
 [![open-quantum-safe](https://circleci.com/gh/open-quantum-safe/oqs-demos.svg?style=svg)](https://app.circleci.com/pipelines/github/open-quantum-safe/oqs-demos)
 
-# oqs-demos
+oqs-demos
+=========
 
 ## Purpose
 
--- a/README.md
+++ b/README.md
@@ -1,7 +1,8 @@
 [![open-quantum-safe](https://circleci.com/gh/open-quantum-safe/oqs-demos.svg?style=svg)](https://app.circleci.com/pipelines/github/open-quantum-safe/oqs-demos)
 
-oqs-demos
-=========
+# oqs-demos
+
+## Purpose
 
 A repository of instructions (with associated patches and scripts) to enable, through [liboqs](https://github.com/open-quantum-safe/liboqs), the use of quantum-safe cryptography in various application software.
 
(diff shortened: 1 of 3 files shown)


xykro:
Ah, there actually **was** a copy+paste error. Thanks for fixing!

baentsch:
> Ah, there actually was a copy+paste error. Thanks for fixing!

Yup - depressing it also slipped my attention :-(

xykro:
> > Ah, there actually was a copy+paste error. Thanks for fixing!
> 
> Yup - depressing it also slipped my attention :-(

For that we have this nice German word "Betriebsblindheit" ;-) Or "Den Wald vor lauter Bäumen nicht sehen" 😄

README.md
```diff
@@ -1,7 +1,8 @@
 [![open-quantum-safe](https://circleci.com/gh/open-quantum-safe/oqs-demos.svg?style=svg)](https://app.circleci.com/pipelines/github/open-quantum-safe/oqs-demos)
 
-oqs-demos
-=========
+# oqs-demos
```
dstebila, line 3: For level 1 and 2 headings, I like using `====` and `----` in Markdown files so that people who read them as plain text files get nice underlines (rather than the semantically equivalent but visually less distinguished `#` and `##` syntax).  (For level 3 headings, I believe `###` is the only supported syntax in Markdown so I use that.)  But this is not worth revising the PR over.

README.md
```diff
@@ -19,3 +20,33 @@ Currently supported packages:
...
+The Open Quantum Safe project is led by [Douglas Stebila](https://www.douglas.stebila.ca/research/) and [Michele Mosca](http://faculty.iqc.uwaterloo.ca/mmosca/)at the University of Waterloo.
+
+## Contributors to oqs-demos include:
+
+    Christian Paquin (Microsoft Research)
+    Dimitris Sikeridis (University of New Mexico / Cisco Systems)
+    Douglas Stebila (University of Waterloo)
+    Goutam Tamvada (University of Waterloo)
+    Michael Baentsch (baentsch.ch)
+    ISE @ FHNW (Fachhochschule Nordwestschweiz)
```
dstebila, line 43: Do you want to put some names or expand out what ISE stands for?

dstebila (approved):

README.md
baentsch: Thanks for letting me know about this convention. Will try to heed it -- for 1st level headings: 2nd levels are hardly anywhere marked with ---. Might make sense to add to https://github.com/open-quantum-safe/liboqs/wiki/Contributing-guide if you'd like to see this generally applied.

Reverted here to maintain convention.

README.md
baentsch: This formatting was intentional: I got feedback that the latest additions were a team effort even though only one person was the github reference; that's why all individuals are listed in the openssh README. ISE @ FHNW written out is a pretty long German term ("Institut für Sensorik und Elektronik, Fachhochschule Nordwestschweiz") I felt breaks the usual length/level of "affiliation detail" at this place. Lastly Google gets you there immediately if interested. Please make a change proposal if you feel this should be different: My first goal is to get this through CI and deployed at Docker Hub and fine-tune later.

PR #72, opened 2021-02-27
fixes #69 [skip ci]

### LINKED ISSUE #69: ERR_CERT_WEAK_KEY in Pre-built Chromium binary
I see a warning about a weak key: ERR_CERT_WEAK_KEY, when connecting to test servers.
pic1
It’s happening with all signature algorithms supported by OQS-enabled Chromium build 0.4.0, except ecdsap256 and rsa3072. The secure connection symbol is missing on server page.
pic2
Also the "subject public key" field in certificate details is empty. 
pic3
Why does Chromium display that field incorrectly? What could be the problem and how to solve it?

baentsch:
Also some doc updates for binary creation and building on Ubuntu 20.


chromium/README.md
```diff
@@ -8,6 +8,8 @@ Further be aware that both cloning the source code as well as building Chromium
 
 2. Navigate to the root directory of the source code, which we will refer to hereafter as `<CHROMIMUM_ROOT>`, and run `git checkout 85.0.4161.2`, which is the latest tag for which we have verified the build instructions. Then, to ensure that all of chromium's third party dependencies are compatible with this tag, run `gclient sync`.
 
+*Note*: Depending on the OS version installed, you may have to install python2, e.g., using `sudo apt install -y python2`, and ensure it's set as the system default, e.g., via `sudo ln -s /usr/bin/python2 /usr/bin/python`.
```
dstebila, line 11: Really? Is it Chromium that uses Python2?  Because I thought all the Python we've been writing has required Python3.

dstebila (approved):

chromium/README.md
baentsch: I also was surprised: Yes, it's the chromium build system (at version 85.0.4161.2) that has this dependency -- it was uncovered when switching to Ubuntu 20 -- which doesn't have python2 by default any more. The cleanest way to get rid of this note is to upgrade to a more current chromium version that shouldn't have this dependency any more (#74)

PR #84, opened 2021-06-02
workaround for upstream issue

baentsch:
Working around https://github.com/openssl/openssl/issues/15569

### DIFF
--- a/openssl3/fulltest/testrun.py
+++ b/openssl3/fulltest/testrun.py
@@ -10,10 +10,11 @@ with open("assignments.json", "r") as f:
    jsoncontents = f.read();
 
 assignments = json.loads(jsoncontents)
+ok=True
 for sig in assignments:
     print("Testing %s:" % (sig))
     for kem in assignments[sig]:
-     if not "_" in kem: # hybrids not yet supported in OSSL3-oqsprovider: TBD
+    # if not "_" in kem: # hybrids not yet supported in OSSL3-oqsprovider: TBD
        # assemble testing command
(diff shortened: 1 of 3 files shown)


openssl3/fulltest/testrun.py
```diff
@@ -10,22 +10,28 @@
    jsoncontents = f.read();
 
 assignments = json.loads(jsoncontents)
+ok=True
 for sig in assignments:
     print("Testing %s:" % (sig))
     for kem in assignments[sig]:
-     if not "_" in kem: # hybrids not yet supported in OSSL3-oqsprovider: TBD
+    # if not "_" in kem: # hybrids not yet supported in OSSL3-oqsprovider: TBD
```
dstebila, line 17: Will this code actually run after line 17 is commented out? Indentation is different once this is commented out, not sure if that's still valid Python.

dstebila (changes requested):

openssl3/fulltest/testrun.py
baentsch: I'd say so -- as long as nobody moves the "if" statement to the same level as the "for".

dstebila: More specifically will the code underneath it still run now that it's at the wrong indentation level?  If you executed the script already then my question can be ignored.

baentsch: Yes, absolutely. You can give it a try with `./testrun.sh openquantumsafe/oqs-ossl3`

baentsch: But actually, I didn't really mean to include this change in the PR :-( All I wanted to change is the ossl3 Dockerfile. So this is an error. PR will be updated, eliminating this.

baentsch:
Replaced by #85

PR #102, opened 2021-09-21
Add scripts to build Chromium with liboqs.

madbrain-vzw:


chromium/scripts/README
```diff
@@ -0,0 +1,28 @@
+These scripts help facilitate a build of PQC-enabled Chromium.
+They have been tested on Ubuntu 20.04 .
+
+How to use :
+
+1. Please set the PROJECT environment variable. This is the location where you
```
baentsch, line 6: If full path required in this variable (relative probably not sufficient?), please state it explicitly.

chromium/scripts/README
```diff
@@ -0,0 +1,28 @@
...
+
+3. If you want the script to run unattended, you may need to increase the
+default sudo timeout. Use 
+sudo visudo
+Then change the timeout value. For example :
+Defaults        timestamp_timeout=360
+
+4. Review getpqc.sh and apply_patch.sh to make sure the Chromium tag and
+corresponding OQS patch are current. If not, please update the files. The
+following lines may need to be updated.
```
baentsch, line 21: Ideally, main README and your script should be kept in line automatically: Would you have a suggestion how to do this? At the least, there should be a link from here to the corresponding chromium version statement in the main README.md.

chromium/scripts/getpqc.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+mkdir -p $PROJECT
+cd $PROJECT
+git clone https://chromium.googlesource.com/chromium/tools/depot_tools.git
+time fetch --nohooks chromium
```
baentsch, line 5: Mights (output from) the `time` commands confuse users?

baentsch (changes requested):
Thanks very much for this contribution, @madbrain-vzw . Beyond the single comments below, one question and one change request:
1) What is the reason these are all small scripts and not one complete script?
2) Could you please also add a line into the main chromium README.md pointing suitably to this so people can find it easily?
Thanks again!

chromium/scripts/README
madbrain-vzw: Agrree. Fix upcoming.

chromium/scripts/README
madbrain-vzw: I'm not certain how to best deal with this. I'm not so familiar with what Markdown offers in terms of variables. I think the simplest solution would be to
1. Isolate the Chromium and patch file name to environment variables in one of the scripts, say set_env.sh . That would be the source of truth.
2. Rename README.md to README.md.template, and have it reference those variables.
3. Finally, run a script invoking set_env.sh and sed to output README.md .
It's about as much work as to change it in two places, but I understand this is better for maintainability not have duplication. If you like the approach, I can do in a future pull request.

chromium/scripts/getpqc.sh
madbrain-vzw: Yes. Will remove it.

madbrain-vzw:
1. This is mainly because I was trying to reproduce the steps documented in README.md . I experienced failures at various steps, and it was helpful for debugging purposes to have this separated.
2. Yes, I will make that change.

chromium/scripts/README
baentsch: > It's about as much work as to change it in two places

Now that you spelled it out I'd conclude that automating this is more work and more convoluted than having to change it in two locations _very rarely_. I'd thus suggest only adding the words "with what is documented in [the README](../README.md])" after "the Chromium tag and corresponding OQS patch are current"

baentsch: Resolved with https://github.com/open-quantum-safe/oqs-demos/pull/102/files/1373a20d28478be44061a18d900baa8e33ea4b43..cb9af78238c9a80d5d89318cb2315b29b356fc28

baentsch (approved):

baentsch:
Thanks very much for this contribution, @madbrain-vzw !

PR #103, opened 2021-10-30
enable use of openssl3 for provider testing

baentsch:
enable use of openssl3 for provider testing

### DIFF
--- a/openssl3/Dockerfile
+++ b/openssl3/Dockerfile
@@ -52,7 +52,7 @@ ENV PATH="${INSTALLDIR}/bin:${PATH}"
 
 # build & install provider (and activate by default)
 WORKDIR /opt/oqs-provider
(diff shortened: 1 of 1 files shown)


baentsch:
@bhess @dstebila This PR is a reminder that `oqsprovider` testing is now tied to the availability of an openssl(3) build tree. Reasons: 1) The [patch](https://github.com/open-quantum-safe/oqs-provider/blob/main/test/ssltestlib.c.patch) to retrofit provider testing into the current openssl(3) master test suite can (and did) break without notice. 2) CI systems anyway always need to build openssl(3) as it is not a system default. 3) This approach enables a more powerful test suite than otherwise doable (without lots of re-inventing the proverbial wheel, i.e., code duplication).
-> Let me know if you are opposed to this approach and want `oqsprovider` to be fully self-contained either by [raising an issue](https://github.com/open-quantum-safe/oqs-provider/issues) or commenting on this thread. I otherwise merge in a day or two.

bhess (commented):
This looks ok to me as it seems only the tests depend on a patched ossl3 build tree.
Just two comments/questions:
Would it make sense to clone a fixed openssl3 tag/commit instead of main to make sure won't break?
Do you know if there is a solution planned in openssl3 that lets providers use these tests without patching the openssl3 base?

baentsch:
> Would it make sense to clone a fixed openssl3 tag/commit instead of main to make sure won't break?

Yes and No: Yes, that would ensure no breakage -- which might be important for a commercial piece of software. But (I decided) No is more important currently, because the provider code should be as close to OSSL3 master as possible: Otherwise, a) (also my) contributions to OSSL3 become available only with a long delay (and/or require new manual tagging and/or even separate provider branches -- which is no effort I'm willing to expend at this time of development); b) breakage makes us aware of OSSL3 changes otherwise going unnoticed -> For as long as no distro makes OSSL3 the default OpenSSL version I thus much prefer a bleeding edge provider to be (failing :) fast. Slow comes later :-)

> Do you know if there is a solution planned in openssl3 that lets providers use these tests without patching the openssl3 base?

That's precisely what I'm doing now (with en/decoder and probably extending to all tests to get rid of the patch) -- at the price of breakage as and when OSSL3 (even test) code changes. But again, I consider this goodness at this time in the OSSL3 lifecycle.

PR #115, opened 2022-01-05
changing the default password

Utopiah:
Was unable to connect using oqs.pw but using `Pa55W0rd` found also in README.md works.


baentsch (approved):
Absolutely correct: Thanks for your contribution!

PR #116, opened 2022-01-07
fix default SIG_ALG from dilithium2 to dilithium3

Utopiah:


baentsch:
Thanks for the catch!

PR #131, opened 2022-02-25
Change nginxquic to nginx.

prchander:
I am not sure if this was supposed to work but openquantumsafe/nginxquic does pull any images. However openquantumsafe/nginx does

### DIFF
--- a/quic/USAGE-server.md
+++ b/quic/USAGE-server.md
@@ -24,7 +24,7 @@ Due to the large number of ports opened the server should be started this way in
 
 ```
    docker network create oqs-quic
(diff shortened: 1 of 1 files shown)


baentsch:
Thanks for making us aware of this mistake. The proposed fix is not OK, though: The correct name is `openquantumsafe/nginx-quic`. The proposed change would pull a non-QUIC-enabled oqs-nginx image. Please try (also interacting with the image) using the oqs-msquic client) for yourself.

baentsch:
Replaced by https://github.com/open-quantum-safe/oqs-demos/commit/f3000ef7a576cf62695fb25c255565b30cbe9e6c

PR #133, opened 2022-03-11
Fix duplicated definition of SIG_ALG variable, Dilithium3 by default.

d1cor:
Remove re-definition of SIG_ALG variable, leaving Dilithium3 as the default signature algorithm, keeping consistency with the README.md file.


baentsch (approved):
Thanks for the contribution!

PR #135, opened 2022-03-29
Adding OpenVPN [skip ci]

baentsch:
CI support not yet added pending review of suitability of image due to
external dependencies.


baentsch:
@dstebila @christianpaquin @crockeea Would you want to glance this over? 

This primarily is meant as an example how to integrate downstream apps with OpenSSL3 and the [oqsprovider](https://github.com/open-quantum-safe/oqs-provider) (no real patch, no fork). Therefore, significant OpenVPN-specific and generic "plumbing" is done by another docker image. OK in your eyes or should we do our own setup scripts (gaining independence from something called [easy-rsa](https://github.com/OpenVPN/easy-rsa) :)?

Feedback also sought as to whether we should add this to the main oqs-demos README and whether to regularly build it via CI. 

By default Kyber is utilized somewhat in line with [Amazon's recent announcement](https://aws.amazon.com/de/about-aws/whats-new/2022/03/aws-kms-acm-support-latest-hybrid-post-quantum-tls-ciphers/) (even though I'm still not sure which exact KEMs are supported there), but of course any other algorithm combination could be chosen: Thoughts on that also elicited. 

@xvzcf @bhess your thoughts would be welcome, too, of course.

openvpn/USAGE.md
```diff
@@ -0,0 +1,62 @@
...
+
+The last three commands clean up all data structures established.
+
+## Advanced usage options
+
+The docker image has been pre-configured to use the quantum-safe crypto (QSC) algorithm family "Kyber" for key establishment but any plain or hybrid QSC algorithm [supported -- see list here](https://github.com/open-quantum-safe/openssl/tree/OQS-OpenSSL_1_1_1-stable#key-exchange) can be selected. To do so, edit the file(s) `/opt/oqssa/bin/serverstart.sh` and `/opt/oqssa/bin/clientstart.sh` suitably as per the comments in those files. Alternatively, change the value for "Groups" in the OpenSSL configuration file `/opt/oqssa/ssl/openssl.cnf` to reflect the QSC KEM algorithm list to be used.
+
+## Disclaimer
+
+[THIS IS NOT FIT FOR PRODUCTIVE USE](https://github.com/open-quantum-safe/openssl#limitations-and-security).
```
dstebila, line 60:
```suggestion
[THIS IS NOT FIT FOR PRODUCTION USE](https://github.com/open-quantum-safe/openssl#limitations-and-security).
```

dstebila:
> This primarily is meant as an example how to integrate downstream apps with OpenSSL3 and the [oqsprovider](https://github.com/open-quantum-safe/oqs-provider) (no real patch, no fork). Therefore, significant OpenVPN-specific and generic "plumbing" is done by another docker image. OK in your eyes or should we do our own setup scripts (gaining independence from something called [easy-rsa](https://github.com/OpenVPN/easy-rsa) :)?

I think it's fine to rely on these other scripts, especially as they are from the OpenVPN team anyway.

> Feedback also sought as to whether we should add this to the main oqs-demos README and whether to regularly build it via CI.

If you think it's sufficiently reliable to advertise as a main demo, then yes, go ahead.

openvpn/USAGE.md
baentsch: So changed.

christianpaquin (approved):

dstebila (approved):

PR #127, opened 2022-02-15
remove msquic platform patch

baentsch:
Update due to https://github.com/microsoft/msquic/pull/2393

PR #138, opened 2022-04-06
fix msquic version

baentsch:
msquic project refactored stuff again causing a breakage in our integration. This PR is to stop the need to keep following them -- at least until someone states a need to have a current oqs-msquic.

### DIFF
--- a/quic/Dockerfile-client
+++ b/quic/Dockerfile-client
@@ -21,7 +21,7 @@ WORKDIR /tmp/merge
 RUN ./merge-oqs-openssl-quic.sh mergeonly
 
 WORKDIR /root
-RUN git clone --depth 1 --branch main https://github.com/open-quantum-safe/liboqs.git && git clone https://github.com/microsoft/msquic && cd msquic && git submodule deinit -f -- submodules/openssl && rm -rf .git/modules/submodules/openssl && git rm -f submodules/openssl && git submodule add /tmp/merge/oqs-openssl-quic submodules/openssl && git submodule update --init submodules/googletest && cd ..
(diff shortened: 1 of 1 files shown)


baentsch:
> @dstebila : Any idea why CI is not running on this PR?

Disregard -- [CCI has an outage](https://status.circleci.com).

christianpaquin (approved):

nibanks:
FYI, I created an entire separate repo for https://github.com/microsoft/quicreach. You might be interested in using it instead of the one that used to be in msquic repo.

baentsch:
> FYI, I created an entire separate repo for https://github.com/microsoft/quicreach. You might be interested in using it instead of the one that used to be in msquic repo.

Yup -- saw it when our build broke again. The decision to fix our integration to a commit when everything still worked was due to the impression that all integration points and APIs are still somewhat "in flux": A few API adaptations back-and-forth (#127 , #130 ) and now even a more drastic refactoring happened... 

I thus opted to not keep following this any more until the dust settles (say a few weeks without further changes). Alternatively, please let me know when you consider things to be/kept stable -- I'll then update things here. Yet another alternative would be for you to take up our [changes](https://github.com/open-quantum-safe/oqs-demos/blob/main/quic/src-tools-reach-reach.cpp.patch) in your code: Strictly speaking they have nothing to do with OQS and just aid our test scripts in checking (and timing) proper QUIC operation.

nibanks:
> Yet another alternative would be for you to take up our [changes](https://github.com/open-quantum-safe/oqs-demos/blob/main/quic/src-tools-reach-reach.cpp.patch) in your code

Sure. I can through together a PR with equivalent changes in the new quicreach.

nibanks:
@baentsch please have a look at https://github.com/microsoft/quicreach/pull/8

PR #159, opened 2022-10-23
Add ngtcp2 to oqs-demos

Keelan10:
Hi, I integrated OQS-OpenSSL-QUIC into ngtcp2 and provided Dockerfiles to fully automate the integration.

Thanks for reviewing!

### DIFF
--- a/ngtcp2/README.md
+++ b/ngtcp2/README.md
@@ -1,8 +1,7 @@
 ## Purpose 
-This directory contains Dockerfiles that build [ngtcp2](https://github.com/ngtcp2/ngtcp2) server and client with  [OQS-OpenSSL-QUIC](https://github.com/open-quantum-safe/oqs-demos/tree/main/quic), which allows ngtcp2 to negotiate quantum-safe keys in TLS 1.3. 
+This directory contains Dockerfiles that build the [ngtcp2](https://github.com/ngtcp2/ngtcp2) server, the ngtcp2 client and quicreach with  [OQS-OpenSSL-QUIC](https://github.com/open-quantum-safe/oqs-demos/tree/main/quic), which allows ngtcp2 to negotiate quantum-safe keys in TLS 1.3. 
 
 
-The client also comes with the [quicreach](https://github.com/microsoft/quicreach) tool, based on a modified version of  [msquic](https://github.com/microsoft/msquic/)  for QSC algorithms.
 
 ## Quick start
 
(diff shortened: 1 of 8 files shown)


ngtcp2/README.md
```diff
@@ -0,0 +1,92 @@
...
+The following commands
+
+```
+docker build -t ngtcp2_client -f Dockerfile-client .
+docker run -it ngtcp2_client bash
+```
+
+will build up and run the container for the QSC-enabled ngtcp2 client.
+
+Ensure that the server and the client run on the same network.
```
baentsch, line 52: Suggest to add code sample using `docker network` like https://github.com/open-quantum-safe/oqs-demos/blob/main/haproxy/USAGE.md

baentsch (changes requested):
Thanks very much for this contribution. This already looks great. Are you interested that this is published on https://hub.docker.com/orgs/openquantumsafe/repositories ? If so, there's two additions needed:
1) A USAGE.md file that could be uploaded as-is to docker hub as documentation for use of the built image. This basically means a split of your current README into a (build-)README and a USAGE.md file. Please see examples in the other demos; of particular relevance might be https://github.com/open-quantum-safe/oqs-demos/blob/main/haproxy/USAGE.md as it also documents docker network issues. 
2) If OK  for you, please also add a suitable entry to the CI system (e.g., along the lines of [the oqs-quic build](https://github.com/open-quantum-safe/oqs-demos/blob/7966f93beae30974872b4321aa290d4d0b882f58/.circleci/config.yml#L102) ).

If you're unfamilar with this, no problem -- I'll then do a followup PR; but please do let us know whether you're OK with this being on docker hub, though.

In any case, please add suitable entries to "package" and "Contributors" to the main [oqs-demos README](https://github.com/open-quantum-safe/oqs-demos/blob/main/README.md).

Keelan10:
Having this published on docker hub would be great. I am okay with that idea.
I would be grateful if you could add an entry to the CI system.

Thanks

baentsch (approved):
LGTM. Thanks for the changes. Will add CI PR.

baentsch:
@Keelan10 While doing the CI integration I hit a problem: The CI-pass condition is to start client and server images automatically and successfully establish a connection. To do that what normally works is assigning the server image a name that then becomes the DNS name under which one can connect. For the client image this works: A server started with `docker run --network ngtcp2-test --name ngtcp2server -it ngtcp-server bash` can be correctly reached from within the client image with the document command `client ngtcp2server 6000`. 

However, I cannot start the server with the equivalent command `server ngtcp2server 6000 CA.key CA.crt` but _have_ to determine its IP address first and then call `server 172.18.0.2 6000 CA.key CA.crt`... -> Is there no way to use the IP address of the server when starting it? The error message I get is 
```
root@31622fcd3db7:~# server ngtcp2server 6000 CA.key CA.crt 
Using document root /root/
getaddrinfo: No address associated with hostname
```

This is odd as `ping ngtcp2server` works just fine:
```
root@31622fcd3db7:~# ping ngtcp2server                    
PING ngtcp2server (172.18.0.2) 56(84) bytes of data.
64 bytes from 31622fcd3db7 (172.18.0.2): icmp_seq=1 ttl=64 time=0.077 ms
64 bytes from 31622fcd3db7 (172.18.0.2): icmp_seq=2 ttl=64 time=0.092 ms
```

Is this easily explicable to you? I didn't want to debug into ngtcp2, really....

Keelan10:
It looks like you found the fix already. 
Thanks for the review. I will use the insight you provided to simplify the usage documentation.

PR #161, opened 2022-10-31
OQS Enabled Envoy Contribution

drouhana:
I hope I have checked all the boxes in the contribution guide.

Thank you!

Daniel


envoy/USAGE.md
```diff
@@ -0,0 +1,25 @@
...
+
+A pre-built [Docker image](https://hub.docker.com/layers/drouhana/envoy-oqs/envoy/images/sha256-e779ccfd8707e31fbf3f47f1f2ac99cb52ea56f6e923a87fbb12b7fa1dbca114?context=repo) has been provided for streamlined use in envoy implementations.
+
+## Local Docker Build
+
+Install [Docker](https://docs.docker.com/get-docker/) and run the following commands:
+
+    docker build -t envoy-oqs .
+
+## Using the Image
```
baentsch, line 15: Only this section should go into USAGE.md. This part would be published on hub.docker.com and should explain the usage of the image. Therefore a bit more explanation of (or a link to other documentation for) "standard envoy images" would be great.

envoy/USAGE.md
```diff
@@ -0,0 +1,25 @@
+## Purpose
```
baentsch, line 1: These 14 initial lines explain docker image building and thus should be (only) part of the README.md.

envoy/Dockerfile
```diff
@@ -0,0 +1,137 @@
...
+# Install bazelisk as Bazel
+WORKDIR /usr/local
+RUN wget --no-check-certificate -O /usr/local/bin/bazel https://github.com/bazelbuild/bazelisk/releases/latest/download/bazelisk-linux-$([ $(uname -m) = "aarch64" ] && echo "arm64" || echo "amd64") \
+    && chmod +x /usr/local/bin/bazel
+
+
+# Build OpenSSL
+WORKDIR /usr/local/openssl-1.1.1g
+RUN ./config \
+    && make -j \
```
baentsch, line 54: Suggest to replace "-j" with $MAKE_DEFINES

envoy/tls/README.md
```diff
@@ -0,0 +1,36 @@
+## Transport Layer Security (TLS) Demo
+
+### Generate Certificates:
+
+All necessary commands are encapsulated in gen_cert.sh. In order to change the encryption algorith, change the term following "-newkey" flag in lines 3 and 5. To execute, navigate to ./certs and run the following command:
```
baentsch, line 5: typo "algorith" (m missing)

baentsch (changes requested):
Thanks for your work and this contribution! Beyond the single comments, one general question: Why is there a need for inclusion of key and certificate material? Could this be eliminated? From a security (and code-scanning) perspective this is not really advisable. Isn't the availability of the "gen_cert.sh" script sufficient? 

Finally, is the understanding correct that you don't want the resultant image(s) to be hosted in the "openquantumsafe" namespace at docker hub? If so, this PR doesn't really need the separate USAGE.md file (as that is what we push to docker hub when publishing an image into "openquantumsafe", i.e., won't need that when not publishing it there). If in turn it would be a goal to publish the image to "openquantumsafe/envoy", there should be a CI script entry added building, testing and pushing the image (see https://github.com/open-quantum-safe/oqs-demos/blob/main/.circleci/config.yml for examples). Lastly, please consider adding a line entry to the main oqs-demos README with your name and the relevant links (see other projects/contributor entries for examples). Thanks again!

baentsch:
@drouhana Would you have time to address the comments above or do my questions/requests raise questions? Please always ask back in such case...

drouhana:
@baentsch thank you for the detailed feedback and changes! I’m sorry for lagging on implementing them, I was catching up on other work and getting over a cold. I will have implemented changes for you in the next 24-48

baentsch:
> I was catching up on other work and getting over being ill. I will have implemented changes for you in the next 24-48

Thanks for the update. Take your time and get well!

drouhana:
@baentsch apologies for dragging on this

- make_define added
- certs removed and instructions for gen_cert.sh added to README in tls folder
- typo fixed
- USAGE.md fixed

> Finally, is the understanding correct that you don't want the resultant image(s) to be hosted in the "openquantumsafe" namespace at docker hub?

Please do add to the openquantumsafe docker! I have updated all lines to reference openquantumsafe/envoy or openquantumsafe/envoy:latest. I assume this doesn't exist yet, so it will throw an error until it does. For example, [here](https://github.com/drouhana/oqs-demos/blob/65cfd957b99b97524c7cc553f07b798d0cbcebac/envoy/tls/Dockerfile#L4)

I have also added to the .circlci. Please let me know if it is terrible and needs correcting, I tried to follow what the other entries had done for testing.

envoy/USAGE.md
baentsch: > images can be used identically to the standard envoy images

Can you please add a link to such documentation? dockerhub doesn't contain info at https://hub.docker.com/r/envoyproxy/envoy

envoy/USAGE.md
baentsch: README.md now states:

> An example implementation of oqs-enabled envoy terminating a tls handshake and proxying to an http backend has been included

Where has this been included? Link(s) would be helpful to understand this statement.

.circleci/config.yml
```diff
@@ -426,6 +463,50 @@ jobs:
...
+          working_directory: envoy
+      - run:
+          name: Build curl with liboqs
+          command: |
+            docker build -t oqs-curl . 
+          working_directory: curl
+      - run:
+          name: Test OQS envoy TLS servers using built OQS envoy
+          command: |
+            ./certs/gen_cert.sh &&
```
baentsch, line 492: This doesn't look right if working directory is set to `envoy/certs`

.circleci/config.yml
```diff
@@ -426,6 +463,50 @@ jobs:
...
+      - setup_remote_docker
+      - run:
+          name: Authenticate to Docker
+          command: echo $DOCKER_PASSWORD | docker login --username $DOCKER_LOGIN --password-stdin
+      - run:
+          name: Build OQS envoy
+          command: |
+            docker build -t envoy-oqs . 
+          working_directory: envoy
+      - run:
```
baentsch, line 484: Is this build "overkill"? Would it be sufficient to simply run `openquantumsafe/curl` and not re-build it? This would safe considerable build time.

.circleci/config.yml
```diff
@@ -426,6 +463,50 @@ jobs:
...
+          name: Build curl with liboqs
+          command: |
+            docker build -t oqs-curl . 
+          working_directory: curl
+      - run:
+          name: Test OQS envoy TLS servers using built OQS envoy
+          command: |
+            ./certs/gen_cert.sh &&
+            docker-compose pull &&
```
baentsch, line 493: `docker-compose` is most likely not available in CircleCI: See documentation [here](https://circleci.com/docs/docker-compose).

baentsch (changes requested):
Thanks for your updates and apologies for reading things as a "naive user": You know everything about envoy (incl. standard usage of the image(s)) but not every reader of the documentation does  -- therefore links to further information would be most welcome to get one started looking at this documentation (or just looking at the docker hub entry to be generated at `openquantumsafe/envoy`).

> Please do add to the openquantumsafe docker! I have updated all lines to reference openquantumsafe/envoy or openquantumsafe/envoy:latest. I assume this doesn't exist yet, so it will throw an error until it does. For example, [here](https://github.com/drouhana/oqs-demos/blob/65cfd957b99b97524c7cc553f07b798d0cbcebac/envoy/tls/Dockerfile#L4)

Does this create a hen-and-egg situation? The CI job must be able to complete (and push the resultant image to dockerhub) before it already is there. Is this guaranteed with this setup (with the second Dockerfile in `tls`)? Apologies if this is clear in your eyes, but I still need to read up on `docker-compose` to understand how you intend it to behave. 

> I have also added to the .circlci. Please let me know if it is terrible and needs correcting, I tried to follow what the other entries had done for testing.

It looks basically OK -- with the caveat that `docker-compose` probably needs to be installed as per the documentation referenced in my comment at the code location.

Finally, did you run the `circle-ci` job added locally on your machine to ensure it runs OK when deploying? See the [how-to](https://circleci.com/docs/local-cli/).

drouhana:
@baentsch here I go again making excuses but... I'm so sorry I will get to this I promise. it is the end of the semester, I'm presenting at a conference next week, and I am interview prepping.

baentsch:
@drouhana Thanks for letting us know. There's definitely more important things:

> I'm presenting at a conference next week, and I am interview prepping.

Good luck for both things! Please post conference material links here if related.

.circleci/config.yml
drouhana: fixed

.circleci/config.yml
drouhana: fixed, you are very correct

.circleci/config.yml
drouhana: noted and fixed. docker-compose install instructions added to config.yml

drouhana:
@baentsch 

I have implemented all your comments to the circleci and tested it locally with no issue! 

Regarding the certs: I have re-added them to the certs folder, as the gen_cert.sh script did not work. It operates on the assumption that the user has openssl-oqs installed onto their local machine. As a result, it is unable to find the openssl-oqs binary within the ubuntu circleci docker image used for testing. 

At that point, my options were either to (1) add another block in the circleci to install openssl-oqs inside the image for key generation, (2) create a docker image executable as a binary that invokes the openssl binary, or (3) re-add the post-quantum certs to the /tls/certs folder

If you have a better solution to the above issue, please let me know. However, the circleci """should""" work now.

Thank you for your patience, I hope you are having a great holiday

baentsch:
> If you have a better solution to the above issue, please let me know. 

What about (also there) using the existing `openquantumsafe/curl` docker image (with virtual mounting) to run `gen_cert.sh` (and removing the hard-coded `openssl` paths in that file)?

drouhana:
That's a great idea! I just implemented it, and it looks like it is working.

I ran into an issue where it was mounting the gen_cert.sh as a directory, then I found this in the circleci documentation after burning a few hours
https://circleci.com/docs/building-docker-images/#mounting-folders

I followed the guidance there and it's working ! Please let me know if there are any issues

.circleci/config.yml
```diff
@@ -455,6 +455,58 @@ jobs:
...
+          no_output_timeout: "1h"
+          command: |
+            docker build -t envoy-oqs . 
+          working_directory: envoy
+      - run:
+          name: Test OQS envoy TLS servers using built OQS envoy
+          command: |
+            docker create -v /data --name certsvolume openquantumsafe/ci-ubuntu-focal-x86_64:latest &&
+            docker cp $(pwd)/certs/gen_cert.sh certsvolume:/data &&
+            docker run -it --user root --volumes-from certsvolume -w /data --name keygen openquantumsafe/curl sh -c "ls -la && /data/gen_cert.sh" &&
```
baentsch, line 491: nit: This `ls` statement seems unnecessary now (?)

baentsch:
> That's a great idea! I just implemented it, and it looks like it is working.
> 
> I ran into an issue where it was mounting the gen_cert.sh as a directory, then I found this in the circleci documentation after burning a few hours https://circleci.com/docs/building-docker-images/#mounting-folders

Sincere apologies for making you waste hours on CCI "quirks": I went through this once myself, so should've warned/pointed you straight to this :-/

The PR now looks very complete and is highly informative for others (incl. myself). 

Just two remaining questions: 
1) What's the purpose of the new, project-global "process.yml" file: Which tooling needs this and for what?
2) Would you want to consider adding reference to the envoy directory (and yourself as contributor) to the main project README.md?

drouhana:
No worries, it was a good learning experience :)

process.yml was only needed for local circleci building and testing, I have removed it and the "ls -la" (good catch, thank you again)

I have added the link and my name to the contributor list as well. Thank you again for your patience, I know I have taken my sweet time finishing this! Please let me know if you need anything else from me for this PR

envoy/USAGE.md
baentsch: @drouhana Did you see the question above?

baentsch: @drouhana : OK to add a link to https://www.envoyproxy.io/docs/envoy/latest/start/docker (as documentation for "standard envoy image" use)? If you don't object I'll do that manually in the docker hub documentation (unless you already add it in this PR/USAGE.md).

baentsch (approved):
Thanks again for the contribution! LGTM now -- only one minor doc link question (sent separately and not a blocker).

envoy/USAGE.md
drouhana: @baentsch Perfectly fine! Thank you very much for doing that

baentsch:
@drouhana The complete Dockerfile is a great contribution to understand how all components are built -- but it just turned out that the build it triggers takes more than the allotted hour (in the CI "free opensource" tier that we use). Do you have an idea how to get around this? How long does the build take for you? On which platform? May a change to [6-hour limited github actions](https://docs.github.com/en/actions/learn-github-actions/usage-limits-billing-and-administration) be an alternative?

envoy/USAGE.md
baentsch: So done: https://hub.docker.com/repository/docker/openquantumsafe/envoy/general

drouhana:
@baentsch oh boy... that's been one of the most annoying things about building envoy from source, it just takes so long. I've built envoy on everything from a local (reasonably fast) machine to an AWS cloud instance, and it's always been painful.

I'm struggling to think of a way to get around this, as the circleci needs to test whether envoy will build with the updated versions of the openssl-oqs suite, right? The envoy [dockerfiles](https://github.com/envoyproxy/envoy/blob/main/ci/Dockerfile-envoy) copy the already built binary, and therefore save space that way. However, this does not help us speed up the actual build-from-source process.

Perhaps we could [change the build command](https://github.com/envoyproxy/envoy/blob/main/bazel/README.md#additional-envoy-build-and-test-options) from `bazel build -c opt envoy` to `bazel build -c fastbuild envoy` ?

baentsch:
> test whether envoy will build with the updated versions of the openssl-oqs suite, right

Yup - that's (would be) the benefit of CI.

How much speedup do you see with this `fastbuild` option? 

Might it be another idea to utilize a docker image built once and only "drop-in" replace the (binary) oqs-components (with their then-current code base) needed by envoy and then run a test? This would unearth at least obviously breaking API and algorithm changes. The only drawback is that this would no longer deliver a cleanly built oqs-envoy at each oqs checkin -- but that's strictly speaking not necessary: There would be an image that could be updated from time to time but we would know that it _could_ be built.

baentsch:
@drouhana Did you see the question above? Any quick thoughts? Sensible or hopeless/too much effort? If the former, could you point to the binaries that need replacement so I could try this if you don't have the time?

PR #167, opened 2022-11-10
Add OpenLiteSpeed to oqs-demos

Keelan10:
Hi, I integrated OQS-BoringSSL into OpenLiteSpeed and provided a Dockerfile to fully automate the integration.

Thanks for reviewing!

### DIFF
--- a/openlitespeed/Dockerfile-server
+++ b/openlitespeed/Dockerfile-server
@@ -27,45 +27,21 @@ RUN cd openlitespeed && git apply --reject --whitespace=fix /opt/build.sh.patch
 RUN cd openlitespeed && ./build.sh &&  sed 's/#OPENLSWS_PASSWORD=/OPENLSWS_PASSWORD=password/' ols.conf  -i && ./install.sh
 
 WORKDIR /root/
-# Generate keys and certificates
-RUN openssl req -x509 \
--new -newkey rsa:3072 \
--keyout CA.key \
--out CA.crt \
--nodes -subj '/CN=oqstest_CA' -days 500
-
-RUN openssl req \
--new -newkey rsa:3072 \
--keyout srv.key \
--out srv.csr \
--nodes \
--subj '/CN= openlitespeed'
-
-RUN openssl x509 -req \
--in srv.csr \
--out srv.crt \
--CA CA.crt \
--CAkey CA.key \
--CAcreateserial \
--extensions v3_req \
--days 365
 
 # Configuring the server to expose CA.crt on port 80 and to enable QUIC on port 443
 COPY httpd_config.conf /usr/local/lsws/conf/
-RUN mkdir -p /var/www/html && cp CA.crt /usr/local/lsws/Example/html/
-
 
 FROM ubuntu:focal 
 
 ENV TZ=Europe/Zurich
 ENV DEBIAN_FRONTEND=noninteractive
 
-COPY --from=builder /root/CA.*  /root/
-COPY --from=builder /root/srv.* /root/
 COPY --from=builder /usr/local/lsws/ /usr/local/lsws
 COPY --from=builder /usr/local/lib/liboqs.* /usr/local/lib
 
 RUN ldconfig
 
-RUN apt update && apt install net-tools
+RUN apt update && apt install -y openssl net-tools
 WORKDIR /root/
+
+COPY serverstart.sh .
--- a/openlitespeed/Dockerfile-server
+++ b/openlitespeed/Dockerfile-server
@@ -1,0 +1,71 @@
+
+FROM ubuntu:focal as builder
+
+ENV TZ=Europe/Zurich
+ENV DEBIAN_FRONTEND=noninteractive
+
+# Update apt software 
+RUN apt update
+
+# Install prerequisites
+RUN apt install git libz-dev libevent-dev cmake gcc ninja-build libunwind-dev pkg-config python3 python3-psutil golang-go -y
+
+WORKDIR /opt
+
+# Download openlitespeed repository
+RUN git clone https://github.com/litespeedtech/openlitespeed.git 
+
+COPY build.sh.patch /opt/
+COPY build_bssl.sh.patch /opt/
+
+# Apply patch to build.sh
+RUN cd openlitespeed && git apply --reject --whitespace=fix /opt/build.sh.patch 
+
+
+# Build and install openlitespeed
+# The webadmin password is set to password
+RUN cd openlitespeed && ./build.sh &&  sed 's/#OPENLSWS_PASSWORD=/OPENLSWS_PASSWORD=password/' ols.conf  -i && ./install.sh
+
+WORKDIR /root/
+# Generate keys and certificates
+RUN openssl req -x509 \
+-new -newkey rsa:3072 \
+-keyout CA.key \
+-out CA.crt \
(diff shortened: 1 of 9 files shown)


openlitespeed/build_bssl.sh.patch
```diff
@@ -0,0 +1,68 @@
...
+ 
+ if [ -d "go" ]; then
+@@ -17,9 +21,9 @@ echo $PATH
+ fi
+ 
+ cd boringssl
+-git reset --hard
+-git checkout master
+-git pull
++# git reset --hard
```
baentsch, line 25: Why is there commented out code? Could this be removed?

openlitespeed/Dockerfile-server
```diff
@@ -0,0 +1,71 @@
...
+
+# Apply patch to build.sh
+RUN cd openlitespeed && git apply --reject --whitespace=fix /opt/build.sh.patch 
+
+
+# Build and install openlitespeed
+# The webadmin password is set to password
+RUN cd openlitespeed && ./build.sh &&  sed 's/#OPENLSWS_PASSWORD=/OPENLSWS_PASSWORD=password/' ols.conf  -i && ./install.sh
+
+WORKDIR /root/
```
baentsch, line 29: Why is there a need to generate certificates (and keys!) and store them in the Docker image? Could this not be done at server startup (avoiding storage of conceptually sensitive material in the docker image)? If I read the client code right, it anyway downloads the CA.crt anew, so that cert/key pair could be generated at server startup, right?

baentsch (changes requested):
Thanks for your contribution. This looks very good, indeed. Some questions to the code separately asked: Please comment. Then, there's reference to a ready-made docker image: For this to be generated, there needs to be code in CCI (.circleci/config.config.yml): Would you be able to add such code? If you're not familiar with CCI, I'll try to do it in a follow-up PR, but maybe you want to give it a try.

openlitespeed/README.md
```diff
@@ -0,0 +1,45 @@
...
+```
+docker build -t lsws -f Dockerfile-server .
+docker network create lsws-test
+docker run --network lsws-test --name lsws -it lsws bash
+```
+
+will run the container for the quantum-safe crypto (QSC) protected OpenLiteSpeed server on the docker network called lsws-test.
+
+### Usage
+Documentation for using the server docker image is contained in the separate [USAGE-server.md](https://github.com/open-quantum-safe/oqs-demos/blob/main/openlitespeed/USAGE-server.md) file.
```
baentsch, line 26: This is a hard coded absolute link: It should be possible to make this relative (so folder name can change in the future and the link also works in your own repo).

openlitespeed/build_bssl.sh.patch
Keelan10: I had commented out those codes as they were no longer relevant.
These unnecessary comments have been removed.

openlitespeed/Dockerfile-server
Keelan10: You are right. I have added a serverstart.sh file which generates the certificate and key files and starts the server.
Thanks!

openlitespeed/README.md
Keelan10: Noted. I made the requested changes.

Keelan10:
> Thanks for your contribution. This looks very good, indeed. Some questions to the code separately asked: Please comment. Then, there's reference to a ready-made docker image: For this to be generated, there needs to be code in CCI (.circleci/config.config.yml): Would you be able to add such code? If you're not familiar with CCI, I'll try to do it in a follow-up PR, but maybe you want to give it a try.

I would be grateful if you could do the CCI code in a follow-up PR.
Thank you!

baentsch:
Thanks for updating the PR as per the other questions. All marked resolved.

> I would be grateful if you could do the CCI code in a follow-up PR. 

Will do my best.  In reviewing the new `serverstart.sh` file there's one thing that makes me wonder, though: Where does the server actually use QSC/QS keys? The certificate seems to be an RSA one and I also didn't find any way to select a QSC KEM: Did I overlook something? Does the [TLS_DEFAULT_GROUPS](https://github.com/open-quantum-safe/oqs-demos/blob/d36a91818d7e12212eb9c094decbd9b31dc1adaf/quic/testrun.py#L25) env var work? Some documentation or a test script (along the lines of https://github.com/open-quantum-safe/oqs-demos/blob/main/quic/testrun.py#L25) would be helpful. It also would help me do the actual CI test script (see sample [here](https://github.com/open-quantum-safe/oqs-demos/blob/d36a91818d7e12212eb9c094decbd9b31dc1adaf/.circleci/config.yml#L103-L136)).

Keelan10:
> Will do my best. In reviewing the new `serverstart.sh` file there's one thing that makes me wonder, though: Where does the server actually use QSC/QS keys? The certificate seems to be an RSA one and I also didn't find any way to select a QSC KEM: Did I overlook something? Does the [TLS_DEFAULT_GROUPS](https://github.com/open-quantum-safe/oqs-demos/blob/d36a91818d7e12212eb9c094decbd9b31dc1adaf/quic/testrun.py#L25) env var work? Some documentation or a test script (along the lines of https://github.com/open-quantum-safe/oqs-demos/blob/main/quic/testrun.py#L25) would be helpful. It also would help me do the actual CI test script (see sample [here](https://github.com/open-quantum-safe/oqs-demos/blob/d36a91818d7e12212eb9c094decbd9b31dc1adaf/.circleci/config.yml#L103-L136)).

I experimented with only QSC KEMs. The server, by default, [supports all the QSC KEMs provided by OQS-BoringSSL](https://github.com/open-quantum-safe/boringssl#key-exchange). I have not found a way to configure the server so that it supports only certain KEMs (instead of supporting all KEMs).

The TLS_DEFAULT_GROUPS env var does work. Please refer to [USAGE-client.md](https://github.com/Keelan10/oqs-demos/blob/openlitespeed/openlitespeed/USAGE-client.md) for performing QSC handshake with the server. I will also add a test script like you requested.

baentsch (approved):

baentsch:
> The TLS_DEFAULT_GROUPS env var does work. Please refer to [USAGE-client.md](https://github.com/Keelan10/oqs-demos/blob/openlitespeed/openlitespeed/USAGE-client.md) for performing QSC handshake with the server. 

Excellent. That documentation is what I was looking for.

> I will also add a test script like you requested

Only as time permits: The PR already is good to merge. Will wait with the merge somewhat so you can add the script if you like.

Thanks again for the contribution!

Keelan10:
> Only as time permits: The PR already is good to merge. Will wait with the merge somewhat so you can add the script if you like.

You may merge it. This will give me time to come up with an extensive test script on a later date.
> Thanks again for the contribution!

Happy to help with OQS!

baentsch:
@Keelan10 : Doing a full run through all algorithms would indeed be good: FYI, when doing "algorithm spot checks", it turns out that this command fails: `SSL_CERT_FILE=CA.crt TLS_DEFAULT_GROUPS=frodo1344shake quicreach lsws --port 443 --stats`. All other KEM algs tested are working OK.... Interesting.... @dstebila , @xvzcf : Are you aware of anything "special" with the combination `frodo1344shake` & OQS-BoringSSL that might cause this "single-algorithm failure"?

PR #168, opened 2022-11-12
adding CI for openlitespeed

baentsch:
@Keelan10 please check whether this looks like a reasonable test condition


Keelan10:
> @Keelan10 please check whether this looks like a reasonable test condition

LGTM

Issue #164, opened 2022-11-09
Wireshark docker image does not recognize PQC algorithm: Display as Unknown


Jpgig:
I have downloaded the Docker Wireshark image as noted on the code page and ran it as per the instructions
I can see the Wireshark version as Version 3.4.9 (Git commit 365e236f5efe) 

I captured the Wireshark traffic but still, it does not recognize the post-quantum crypto algorithm in the handshake process.
![image](https://user-images.githubusercontent.com/115164852/200841197-f8289e97-483e-4419-b6dc-ce7db669caaa.png)

anhu:
These are for signature algorithms.  When I first did this wireshark modification, we didn't yet support signature schemes and so I did not think to make the modification for them.  

@Jpgig , as a workaround, you can use the oqs style names.  I think those should be populated.

Jpgig:
Hi Anthony, I also used oqs style names but still, it does not populate the PQC  algo name. Can you suggest any other place where I can look for the changes?

anhu:
hi @Jpgig , I'll have to look into this more deeper. You'll have to give me a bit of time as I'm currently looking into other matters. 
Please check for updates by end of next week. 
Warm regards, Anthony

anhu:
Hi @Jpgig , 

I just check and noticed that I had not updated for Dilithium as when we did this integration, we did not yet support Dilithium.   Please see  https://github.com/open-quantum-safe/oqs-demos/pull/170 which updates our integration.

anhu:
@Jpgig , 

Please try again. I've made sure it works: 

![image](https://user-images.githubusercontent.com/1266573/203428592-7a21753d-c4f6-4208-9bfe-dea3ab2ec60e.png)

If docker doesn't seem to work, try using build.sh.  Make sure to change the flavour at the top. 

Warm regards, Anthony

anhu:
to try the build.sh: 

cd oqs-demos/wireshark
git fetch 
git checkout origin/wireshark_wolf_dilithium
# edit build.sh to change to wolfssl flavor
./build.sh
# enter your sudo password
# you might have problems with software that needs to be install.... resolve them. 
cd wireshark-3.4.9/build 
sudo ./run/wireshark &
# enter your sudo password
```
You should know what to do now is specific to your situation. 

Warm regards, Anthony

Jpgig:
Thanks, Anthony
I used all updated files but still it doe not identify the Signature algorithm in Wireshark.
[...]

PR #170, opened 2022-11-21
Adding code points and OIDs for Dilithium.

anhu:


anhu:
@baentsch , can you please merge at your earliest convenience?

wireshark/wolfssl-qsc.h
```diff
@@ -31,6 +37,12 @@
     { 12094, "p256_kyber_90s_level1" }, \
     { 12095, "p384_kyber_90s_level3" }, \
     { 12096, "p521_kyber_90s_level5" }, \
-    { 0xfe0b, "falcon_level1" },\
-    { 0xfe0e, "falcon_level5" },
+    { 0xfe0b, "falcon_level1" }, \
+    { 0xfe0e, "falcon_level5" }, \
+    { 0xfea0, "dilithium_level2" }, \
```
baentsch, line 42: SIGs under the "QSC_KEMS" define seem "weird": Is this intentional?

baentsch (changes requested):
> @baentsch , can you please merge at your earliest convenience?

Before doing this, please confirm that your fix should only change the WolfSSL analysis behaviour and not the OQS analysis behaviour (which is default for the Dockerfile). Thus, this fix may not be complete.

wireshark/wolfssl-qsc.h
anhu: Yes, I would agree, but this was the convention that you created.  (Falcon was already there in PR 105 which I believe you authored).  Probably better names would be `QSC_CODEPOINTS` and `QSC_OIDS` but that is out of scope of this change.

anhu: Hi @baentsch , 

Sorry for the delayed response on the comment below, had to go eat lunch.  

I prefer you to handle the OQS side.  The reason being you decide your naming conventions and which names are to be displayed or left unknown for your flavour.

anhu:
> Before doing this, please confirm that your fix should only change the WolfSSL analysis...

Confirmed.  I do not want to speak for the OQS Naming conventions and what is left unknown.

anhu:
@baentsch , I have a fix, but it will require some help from you as my `sed` is not as good as yours! 

```
I need the following change added to `epan/dissectors/packet-tls-utils.c` : 
/* RFC 8446 Section 4.2.3 */
const value_string tls13_signature_algorithm[] = {
    { 0x0201, "rsa_pkcs1_sha1" },
    { 0x0203, "ecdsa_sha1" },
    { 0x0401, "rsa_pkcs1_sha256" },
    { 0x0403, "ecdsa_secp256r1_sha256" },
    { 0x0501, "rsa_pkcs1_sha384" },
    { 0x0503, "ecdsa_secp384r1_sha384" },
    { 0x0601, "rsa_pkcs1_sha512" },
    { 0x0603, "ecdsa_secp521r1_sha512" },
    { 0x0804, "rsa_pss_rsae_sha256" },
    { 0x0805, "rsa_pss_rsae_sha384" },
    { 0x0806, "rsa_pss_rsae_sha512" },
    { 0x0807, "ed25519" },
    { 0x0808, "ed448" },
    { 0x0809, "rsa_pss_pss_sha256" },
    { 0x080a, "rsa_pss_pss_sha384" },
    { 0x080b, "rsa_pss_pss_sha512" },
QSC_KEMS
    { 0, NULL }
};
```
I could probably figure it, but its my EOD.  Let me know if you are able to help.

anhu:
To be more explicit, QSC_KEMS needs to be added to tls13_signature_algorithm .

anhu:
The sed wasn't as bad as I thought.  External 3rd party confirmed the fix.

baentsch:
> The sed wasn't as bad as I thought. 

Good to hear that and sorry I currently don't have the bandwidth to set this up myself -- just too many elements involved costing too much time to confirm correct operation :-(

> External 3rd party confirmed the fix.

Which confirmation are you referring to? I only found [this](https://github.com/open-quantum-safe/oqs-demos/issues/164#issuecomment-1330749077) -- and it doesn't look OK. Also, does this confirmation refer to the "wolfssl" flavour or the "OQS" flavour -- which is the default here? I'm still surprised seeing the define `OQS_KEM` used to carry signature names: That does not look right: KEMs and SIGs are clearly separated [in the OQS master file here](https://github.com/open-quantum-safe/openssl/blob/OQS-OpenSSL_1_1_1-stable/qsc.h). I do understand that this approach can work for signatures in the "manipulated" `wolfssl-qsc.h" file but not for the "OQS" version: Hence the question: What confirmation (using which setup) has been given?

Jpgig:
Thanks, Anhu for providing a private git branch path. That worked correctly for me.

anhu:
The private branch @Jpgig is referring to is the branch for this PR: ` wireshark_wolf_dilithium` 

> What confirmation (using which setup) has been given?

This was done using `build.sh` under the `wolfssl` flavour.

baentsch (approved):

baentsch:
Merging as a stopgap fix for wolfssl. Will follow up with a longterm fix. Will let you know when done @anhu so you could update the wolfssl-specifics if you want.

anhu:
@baentsch , Thank you very much.  I look forward to your update.

baentsch:
@anhu so now done

anhu:
Excellent. Thank you so much @baentsch !!

PR #175, opened 2022-12-15
remove firesaber from openlitespeed CI test

baentsch:
Correct algorithm choice after https://github.com/open-quantum-safe/boringssl/pull/88 landed

As operational/integration error emerged doing so, disable CCI support for openlitespeed
As timing problems emerged, disable envoy CCI support


baentsch:
@Keelan10 If there's no feedback to the contrary from you, I'll have to completely disable in this PR openlitespeed from the CI process due to the problem documented in https://github.com/open-quantum-safe/oqs-demos/pull/167#issuecomment-1383673300

dstebila (approved):

Issue #185, opened 2023-03-01
OpenSSH demo fails to connect


psschwei:
I get an `ssh: connect to host localhost port 2222: Address not available` error when trying to run the openssh demo:

```console
$ docker version
Client: Docker Engine - Community
 Version:           23.0.1
 API version:       1.42
 Go version:        go1.19.5
 Git commit:        a5ee5b1
 Built:             Thu Feb  9 19:50:04 2023
 OS/Arch:           linux/amd64
 Context:           default

Server: Docker Engine - Community
 Engine:
  Version:          23.0.1
  API version:      1.42 (minimum version 1.12)
  Go version:       go1.19.5
  Git commit:       bc3805a
  Built:            Thu Feb  9 19:47:02 2023
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.6.18
  GitCommit:        2456e983eb9e37e47538f59ea18f2043c9a73640
 runc:
  Version:          1.1.1
  GitCommit:        v1.1.1-0-g52de29d7
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0

$  docker run --name oqs-openssh-server -dit --rm openquantumsafe/openssh
74b0fe3f841ab2ea0b15cb1b55404360bc45669b74fa59c529b8a9077d951325

$ docker ps
CONTAINER ID   IMAGE                     COMMAND                  CREATED          STATUS          PORTS      NAMES
74b0fe3f841a   openquantumsafe/openssh   "./entrypoint.sh /bi…"   10 seconds ago   Up 10 seconds   2222/tcp   oqs-openssh-server

$  docker exec --user oqs -it oqs-openssh-server ssh oqs@localhost
ssh: connect to host localhost port 2222: Address not available
```

Though the above is using the image hosted on DockerHub, I also get the same error when building my own image via `docker build -t oqs-openssh-img .`

psschwei:
The "Connect from container to container via OQS-SSH" section worked, though it took ~15 minutes to start oqs-sshd... will retry the quick start and give it a similar amount of time to get ready...

psschwei:
Yes, that was the issue, just needed to wait a bit for the container to be ready.

dstebila:
Taking 15 minutes to start does seem like a problem.  It might be related to key generation -- on first execution the server might be generating keys, and some of the keys take a while to generate.  Not entirely sure.  Any insight, @christianpaquin or @baentsch ?

[...]

PR #187, opened 2023-03-02
Change openssh default host key algo to dilithium

psschwei:
Follow up from discussion in https://github.com/open-quantum-safe/oqs-demos/issues/185#issuecomment-1451942190 , where changing the default host key algo to dilithium was mentioned.

### DIFF
--- a/openssh/ssh_config
+++ b/openssh/ssh_config
@@ -49,10 +49,10 @@ Port 2222
 #IdentityFile ~/.ssh/id_ssh-picnicL3FS
 #IdentityFile ~/.ssh/id_ssh-sphincsharaka128fsimple
 
-#IdentityFile ~/.ssh/id_ssh-ecdsa-nistp256-dilithium2aes
+IdentityFile ~/.ssh/id_ssh-ecdsa-nistp256-dilithium2aes
 #IdentityFile ~/.ssh/id_ssh-ecdsa-nistp256-falcon512
 #IdentityFile ~/.ssh/id_ssh-ecdsa-nistp256-picnicL1full
-IdentityFile ~/.ssh/id_ssh-ecdsa-nistp384-picnicL3FS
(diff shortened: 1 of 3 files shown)


baentsch (changes requested):
This would ne switching to dilithium2aes, a pretty weak and non NIST-approved algorithm. Suggest using p384_dilithium3 instead.

openssh/ssh_config
```diff
@@ -49,10 +49,10 @@ Port 2222
 #IdentityFile ~/.ssh/id_ssh-picnicL3FS
 #IdentityFile ~/.ssh/id_ssh-sphincsharaka128fsimple
 
-#IdentityFile ~/.ssh/id_ssh-ecdsa-nistp256-dilithium2aes
+IdentityFile ~/.ssh/id_ssh-ecdsa-nistp256-dilithium2aes
 #IdentityFile ~/.ssh/id_ssh-ecdsa-nistp256-falcon512
 #IdentityFile ~/.ssh/id_ssh-ecdsa-nistp256-picnicL1full
-IdentityFile ~/.ssh/id_ssh-ecdsa-nistp384-picnicL3FS
+#IdentityFile ~/.ssh/id_ssh-ecdsa-nistp384-picnicL3FS
```
baentsch, line 55: Also, any reference to picnic should be removed to not tempt users into activating a non-supported algorithm again.

baentsch: Thanks for purging picnic from the USAGE file too. To complete things, could you please also kick it from all config files? Would save me a PR just for that...

baentsch (approved):

openssh/ssh_config
psschwei: my bad, I misinterpreted what you meant there. will fix now

baentsch (approved):

PR #189, opened 2023-03-14
Update Chromium build instructions

Raytonne:
+ Add instructions to build Chromium on Windows
+ Add prebuilt 64-bit Windows Chromium installer


baentsch:
In the light of https://github.com/open-quantum-safe/boringssl/pull/100, suggest to either move this PR to "Draft" (and re-base) or close until `oqs-boringssl` is ready.

Issue #182, opened 2023-02-25
replace oqs-openssl111


baentsch:
With https://github.com/openssl/openssl/pull/19312 merged, oqs-provider together with OpenSSL3 (master) now deliver the same level of functionality as oqs-openssl111. 
This issue is to propose replacing oqs-openssl111 with openssl3+oqs-provider where possible in the demos.

Applicable integrations (tick if done) -- suggested order of importance
- [x] openssl3
- [x] OpenVPN
- [x] curl
- [x] nginx
- [x] httpd
- [x] epiphany
- [ ] envoy
- [x] quic
- [ ] mosquitto
- [x] ngtcp2
- [ ] unbound
~- [ ] haproxy~ (available [upstream](https://github.com/haproxy/wiki/wiki/OpenSSL-Providers-in-HAProxy#oqsprovider---open-quantum-safe-provider))
- [ ] [liboqs-python](https://github.com/open-quantum-safe/liboqs-python/blob/main/docker/Dockerfile)

Not applicable: Wireshark, Chromium, openssh, openlitespeed

baentsch:
`haproxy` testing (via curl) now fails as `haproxy` is still using `oqs-openssl111` and we severed the "interoperability tie" between `oqs-provider` (used by `curl`) and `oqs-openssl111` (used by `haproxy`).

Question thus: Would anyone mind we drop `haproxy` from the list of supported (and tested) integrations (until someone finds time and interest again to support it -- via new PR)?

baentsch:
Given today's decision to keep supporting oqs-openssl111 work on this topic is put on the backburner. I personally would very much welcome other's contributions regarding maintenance of oqs-openssl111.

baentsch:
@dr7ana @igorbarshteyn @Keelan10 @chiachin2686 @ryndia  You all kindly contributed oqs-openssl111 integrations to `oqs-demos` and we'd like to ask whether you'd also be willing to help move these to `openssl`v3.

[...]

PR #190, opened 2023-03-16
upgrade curl and move to openssl3

baentsch:
A further step towards #182

### DIFF
--- a/curl/Dockerfile
+++ b/curl/Dockerfile
@@ -1,7 +1,7 @@
 # Multi-stage build: First the full builder image:
 
 # define the Curl version to be baked in
-ARG CURL_VERSION=7.73.0
+ARG CURL_VERSION=7.87.0
 
 # Default location where all binaries wind up:
 ARG INSTALLDIR=/opt/oqssa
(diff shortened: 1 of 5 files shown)


curl/Dockerfile
```diff
@@ -39,29 +39,31 @@ RUN apk add build-base linux-headers \
...
+# set path to use 'new' openssl. Dyn libs have been properly linked in to match
 ENV PATH="${INSTALLDIR}/bin:${PATH}"
 
+# build & install provider (and activate by default)
+WORKDIR /opt/oqs-provider
+RUN ln -s ../openssl . && cmake -DOPENSSL_ROOT_DIR=${INSTALLDIR} -DCMAKE_BUILD_TYPE=Release -DCMAKE_PREFIX_PATH=${INSTALLDIR} -S . -B _build && cmake --build _build  && cp _build/lib/oqsprovider.so ${INSTALLDIR}/lib64/ossl-modules && sed -i "s/default = default_sect/default = default_sect\noqsprovider = oqsprovider_sect/g" /opt/oqssa/ssl/openssl.cnf && sed -i "s/\[default_sect\]/\[default_sect\]\nactivate = 1\n\[oqsprovider_sect\]\nactivate = 1\n/g" /opt/oqssa/ssl/openssl.cnf && sed -i "s/providers = provider_sect/providers = provider_sect\nssl_conf = ssl_sect\n\n\[ssl_sect\]\nsystem_default = system_default_sect\n\n\[system_default_sect\]\nGroups = \$ENV\:\:KEM_ALG\n/g" /opt/oqssa/ssl/openssl.cnf && sed -i "s/\# Use this in order to automatically load providers/\# Set default KEM alg if not set via environment variable\nKEM_ALG = kyber512\n\n# Use this in order to automatically load providers/g" /opt/oqssa/ssl/openssl.cnf
```
dstebila, line 65: Yikes!

baentsch: I know, I know. This is primarily meant to protect from "unforeseen" upstream changes to the default `openssl.cnf` -- which I consider a risk as we always build against `master`. We could achieve this also with `patch` if you prefer that (?) -- but might have to update more often then.

dstebila: No need to patch, just commenting on it.  Go ahead with merging as ready.

PR #194, opened 2023-04-12
Update httpd to OpenSSL3

baentsch:
PR done such as to easily apply after #191 gets merged.

CI problem remaining: OpenSSH doesn't pass any more: @christianpaquin @dstebila @xvzcf : Could one of you maybe take a look as to what's the (apparently new) problem with that? Otherwise, I'd suggest also removing `openssh` from the supported demos list just like `haproxy` until someone has time & interest to fully support it again.


baentsch:
Thanks @xvzcf for the "naming updates" & sorry I forgot about them.

dstebila (approved):

baentsch:
Same here: I'm merge this as-is right away, but won't be around if sth breaks in consequence; thus, leave merging to you @dstebila @xvzcf

PR #195, opened 2023-04-12
Size optimization ngtcp2 and change to use  OpenSSL3/oqsprovider 

Keelan10:
Hi,
I have refactored the Dockerfiles for ngtcp2 to use multi-stage build. This reduces the size of each docker image to ∼32 MB.

Thank you for reviewing!

Edit: ngtcp2 has been moved to openssl3.


baentsch:
Thanks for this PR and optimization, @Keelan10 . Unfortunately, I'm heading out to vacation in a few minutes, so can I ask @xvzcf  to take a look? Please also consider moving this to use OpenSSL3+oqsprovider as per #182.

Keelan10:
Understood. Have a great vacation!

ngtcp2/Dockerfile-client
```diff
@@ -1,59 +1,71 @@
...
+
+# Copy all required shared object dependencies to a single directory
+RUN mkdir /opt/lib && cd /opt/lib && \
+        cp /opt/ngtcp2/lib/.libs/libngtcp2.so.11 . && \
+        cp /usr/lib/libev.so.4 . && \
+        cp /opt/nghttp3/build/lib/libnghttp3.so.5 . && \
+        cp /opt/ngtcp2/crypto/openssl/.libs/libngtcp2_crypto_openssl.so.4 . && \
+        cp /opt/oqssa/lib64/libssl.so.81.3 . && \
+        cp /opt/oqssa/lib64/libcrypto.so.81.3 . && \
+        cp /usr/lib/libstdc++.so.6 . && \
```
baentsch, line 56: These `cp` commands look "brittle": What if any of the library versions change? The docker build will not succeed any more...

baentsch (changes requested):
Apologies for the long silence on our side :-( and Thanks very much again for this PR -- particularly for the update to OpenSSL3/removing the dependency on oqs-openssl111!

In general, I'm happy with this PR, but when running it locally, the build failed, so CI will also fail :-( So, can I ask you to re-run the build along the lines of what the CI logic does and update the code as necessary? I'm particularly concerned about the hard-coded library version copies (see single comment ). For the actual commands that need to succeed for CI see https://github.com/open-quantum-safe/oqs-demos/blob/9a9b35fc28875e1fbe3ec1b6e6a8d4f0ca93b633/.circleci/config.yml#L406-L414)? In addition/alternatively, please consider activating a CircleCI account to run the CI under your account ID: We'd then be sure everything is OK prior to merge. If you don't like CircleCI, please consider using GitHub CI instead.

ngtcp2/Dockerfile-client
Keelan10: Thank you for the feedback! I changed them to use wildcards instead.

Keelan10:
> In general, I'm happy with this PR, but when running it locally, the build failed, so CI will also fail :-( 

It turns out the build was failing due to changes in the library versions. It builds properly now that I have addressed this by using wildcards. Once again thank you for your valuable input!

baentsch (changes requested):
> It turns out the build was failing due to changes in the library versions. It builds properly now that I have addressed this by using wildcards. Once again thank you for your valuable input!

Thank you for the update! Everything now indeed builds OK. However, the CI test (still) fails (see command sequence above): For one, the new docker image doesn't contain bash, but secondly, no (test) CA is "baked into" the image: How would you recommend to change things to have a good CI test for this PR?

baentsch:
@Keelan10 I added your PR code to our CI chain and various failures ensued: If you still would like to see this PR move forward, would you please re-base your code such as to get past the non-PR related CI failures (of course with the goal to  see that your new code also passes CI)?

baentsch (approved):

Keelan10:
Hello, 
I apologize for the late reply. I have rebased the branch as requested. Please let me know if the CI encounters any failure due to  this PR.

baentsch:
Superseded/replaced by #211

PR #196, opened 2023-04-12
Add h2load to oqs-demos

Keelan10:
Hi, I integrated openssl v3 and oqs-provider into h2load benchmarking tool and provided a Dockerfile to fully automate the integration.

Thanks for reviewing!


baentsch:
Thanks again for the contribution but the same comment here: We really want to reduce the number of demos using the old `oqs-openssl111` based code as per https://github.com/open-quantum-safe/openssl/discussions/434. If you'd have the time to move things over to use OpenSSL3+oqsprovider that'd be great. Again tagging @xvzcf for attention during my absence.

baentsch:
> I don't think this is necessarily a blocker, what do you think @baentsch?

If it's only this one algorithm (failing) I'd agree. Is this the case?

xvzcf:
> If it's only this one algorithm (failing) I'd agree. Is this the case?

`frodo1344shake` probably fails because of the large artifacts it produces, but yes @Keelan10 could you confirm what algorithms work and what don't?

Keelan10:
> `frodo1344shake` probably fails because of the large artifacts it produces, but yes @Keelan10 could you confirm what algorithms work and what don't?

Here is a list of algorithms that succeed and fail:
Succeed:
- bikel1
- kyber512, kyber768, kyber1024, kyber90s512, kyber90s768, kyber90s1024
- frodo640aes, frodo640shake, frodo976shake
- hqc128, hqc192

Fail:
- bikel3, bikel5
- frodo976aes, frodo1344aes, frodo1344shake
- hqc256

baentsch:
Thanks very much for creating this list, @Keelan10 ! This is the first time in a long time that I'm seeing such big differences between the different algorithms. This really looks worth while a deeper investigation. Do you agree @xvzcf @dstebila ? @Keelan10 Can you share (add to the PR) a script (ideally, to be made part of CI eventually) that creates this result?

xvzcf:
> Thanks very much for creating this list, @Keelan10 ! This is the first time in a long time that I'm seeing such big differences between the different algorithms. This really looks worth while a deeper investigation. Do you agree @xvzcf @dstebila ? @Keelan10 Can you share (add to the PR) a script (ideally, to be made part of CI eventually) that creates this result?

I think it is worth a closer look, probably has to do with the sizes of the keys/ciphertexts produced by those algorithms, @Keelan10 could you try debugging the segfault with frodo1344shake and seeing if a fix for that fixes it for other algorithms as well?

dstebila:
Sizes are a good theory, although the communication (input/output) size of frodo976aes and frodo976shake are identical, so why would one fail but not the other?  Maybe something involving memory usage as well?

xvzcf:
> Sizes are a good theory, although the communication (input/output) size of frodo976aes and frodo976shake are identical, so why would one fail but not the other? Maybe something involving memory usage as well?

Ah true, and yeah it could be

baentsch (approved):
Thank you very much for this PR -- and particularly the OpenSSL3 integration: It works like a charm! Also Thanks for the script to trigger all algorithms (and the various Segmentation faults as per the discussion). I'd be fine to merge this as-is -- only tiny improvement suggestion: Would it be reasonable to extend the list of algorithms in "check_algorithms.sh" with the hybrid variants? Please let us know whether you'd like to update this before merge.

xvzcf:
I think it'd be nice to figure out what the source of the segfaults is before we merge it in, if only to preclude people posting issues about failures they encounter.

baentsch:
> I think it'd be nice to figure out what the source of the segfaults is before we merge it in, if only to preclude people posting issues about failures they encounter.

Nice, yes, but does anyone of us have time for this? We're already a month late with v0.8.0... 

So, I'd be willing to take the risk of such questions. If a single one comes, I'd be surprised -- and add a note to the README :)

xvzcf (approved):

baentsch:
Thanks for the contribution, @Keelan10 !

PR #199, opened 2023-05-11
Update README.md to fix the cable formatting

Utopiah:
HAPproxy commented line broken the following ones


baentsch (approved):
Indeed. Looks much better this way. And given that no-one complained about the removal of the `haproxy` demo, I'm fine with this.

PR #204, opened 2023-05-28
Adds docker build to generate test server package

bhess:
Adds docker-based build for test server package, using ubuntu docker base.

Difference to previous test server config: only (pq-)signatures and kems with matching security levels are configured (exception are the baseline rsa/ecdsa signatures).

Note: The PR currently includes a common.py, which can be automatically loaded once https://github.com/open-quantum-safe/oqs-provider/pull/173 is merged.

### DIFF
--- a/nginx/fulltest-provider/common.py
+++ b/nginx/fulltest-provider/common.py
@@ -3,25 +3,7 @@ import subprocess
 import pathlib
 import psutil
 import time
-
-key_exchanges = [
-##### OQS_TEMPLATE_FRAGMENT_KEX_ALGS_START
-    # quantum-safe key exchanges
-    ('frodo640aes', 128),('frodo640shake', 128),('frodo976aes', 192),('frodo976shake', 192),('frodo1344aes', 256),('frodo1344shake', 256),('kyber512', 128),('kyber768', 192),('kyber1024', 256),('bikel1', 128),('bikel3', 192),('bikel5', 256),('hqc128', 128),('hqc192', 192),('hqc256', 256),
-    # quantum-safe + classical NIST key exchanges
-    ('p256_frodo640aes', 128),('p256_frodo640shake', 128),('p384_frodo976aes', 192),('p384_frodo976shake', 192),('p521_frodo1344aes', 256),('p521_frodo1344shake', 256),('p256_kyber512', 128),('p384_kyber768', 192),('p521_kyber1024', 256),('p256_bikel1', 128),('p384_bikel3', 192),('p521_bikel5', 256),('p256_hqc128', 128),('p384_hqc192', 192),('p521_hqc256', 256),
-    # quantum-safe + classical X key exchanges
-    ('x25519_frodo640aes', 128),('x25519_frodo640shake', 128),('x448_frodo976aes', 192),('x448_frodo976shake', 192),('x25519_kyber512', 128),('x448_kyber768', 192),('x25519_bikel1', 128),('x448_bikel3', 192),('x25519_hqc128', 128),('x448_hqc192', 192),    
-##### OQS_TEMPLATE_FRAGMENT_KEX_ALGS_END
-]
-signatures = [
-    ('ecdsap256', 0), ('rsa3072', 0),
-##### OQS_TEMPLATE_FRAGMENT_SIG_ALGS_START    # quantum-safe signatures
-    ('dilithium2', 128),('dilithium3', 192),('dilithium5', 256),('falcon512', 128),('falcon1024', 256),('sphincssha2128fsimple', 128),('sphincssha2128ssimple', 128),('sphincssha2192fsimple', 192),('sphincsshake128fsimple', 128),
-    # quantum-safe + classical signatures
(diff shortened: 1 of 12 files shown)


nginx/fulltest-provider/common.py
```diff
@@ -0,0 +1,154 @@
+import os
+import subprocess
+import pathlib
+import psutil
+import time
+
+key_exchanges = [
+##### OQS_TEMPLATE_FRAGMENT_KEX_ALGS_START
```
baentsch, line 8: Really confused by this: We're running `jinja2`in oqs-demos? Which script does this? I'd _really_ refrain from this complication. Why not query openssl for the algorithms?

baentsch: Ahh -- now I get it: You copied this file over from https://github.com/open-quantum-safe/oqs-provider/pull/173, right? IMO this is confusing: Please at least do it as before and pull it over via `wget` (such that one knows where it's coming from -- and that it's somewhat current as `jinja2` only runs in `oqs-provider`) -- or (my strong preference) please do it so that it extracts algorithms from running `openssl list` and thus adapts fully to the then-current algorithm capabilities in `oqs-provider` (whichever binary one chooses to run).

baentsch (changes requested):
See separate comment

nginx/fulltest-provider/common.py
bhess: > Ahh -- now I get it: You copied this file over from [open-quantum-safe/oqs-provider#173](https://github.com/open-quantum-safe/oqs-provider/pull/173), right? IMO this is confusing: Please at least do it as before and pull it over via `wget` (such that one knows where it's coming from -- and that it's somewhat current as `jinja2` only runs in `oqs-provider`)

Yes exactly, I intended to uncomment https://github.com/open-quantum-safe/oqs-demos/blob/ed50ec659841b4a26a02fcdc6ccb7abe0b62eb84/nginx/fulltest-provider/build_ubuntu.sh#L5-L6
once the script is merged in oqs-provider, and then delete it here. Wget'ting it now from the PR in oqs-provider would lead to a stale link once the PR is merged (or not).

> or (my strong preference) please do it so that it extracts algorithms from running `openssl list` and thus adapts fully to the then-current algorithm capabilities in `oqs-provider` (whichever binary one chooses to run).

I also need the security strengths to be able to generate a config that only consists of sigs and kems with matching strengths. I don't think this info is available with `openssl list`, or is it? Would be great if it is. If not, you see a way to get this info without jinja templating?

baentsch: > I don't think this info is available with openssl list, or is it?

It is: https://github.com/open-quantum-safe/oqs-provider/pull/156

bhess: Oh, that's nice, thank you for the pointer! This really seems the better solution.

I'd then change this to use `openssl list` instead of jinja templating, remove the jinja-generated parts from the common.py and keep the rest of the script here. So https://github.com/open-quantum-safe/oqs-provider/pull/173 can be closed as well.

It will probably take me a few days to come back to this here.

bhess: The PR is updated with a commit that removes jinja-generated parts and directly gets the algorithms, including their security strengths with `openssl list`.

baentsch (commented):
Thanks for de-jinja-ing this. Not sure I am enthused about seeing scala appear instead :-} But then again, this code 
- does not seem to be the same running at test.openquantumsafe.org -- at least this PR contains lots of references to Chromium and the test server only contains one line (intentional, btw?),
- is not used for CI,
- is not pushed to docker hub, 

so I don't really mind. 
This begs the question, though: Is there any automation using this code or is it purely for demo purposes and your internal use?

baentsch (approved):

bhess:
Deployment isn't fully automated as the tar archive generated by the docker build still needs to be copied to the server. The docker build is however the same running at test.openquantumsafe.org - Chromium-related pages are generated but not referenced (keeping them around just in case there is Chromium support again in the future).

baentsch:
> Chromium-related pages are generated but not referenced

Hmm:

image

bhess:
With Chromium-related page, I meant the dedicated page on test.openquantumsafe.org that lists all supported ports.

> Hmm:
> 
> image

I left this in intentionally, for awareness of the Chromium limitations, without advertising specific ports. So would you remove this as well?

baentsch:
> for awareness of the Chromium limitations, without advertising specific ports. So would you remove this as well?

I personally would: We really don't support Chromium --or boringssl-- any more, right? Besides, the support always was very limited (only a few algorithm combinations -- hence the text you left in). Therefore, I would go even further and would remove as part of this PR Chromium from the list of supported demos -- I did the same with `haproxy` as there was no feedback to the opposite (no-one speaking up voicing interest to use it). Code remained, but no CI any more. That way, it can be resuscitated if someone would pledge the time to maintain it. 

@dstebila @christianpaquin What's your take? We could then even close #189 as there was no activity on this from our side (@Raytonne, please speak up if you see a need for this).

dstebila:
> > for awareness of the Chromium limitations, without advertising specific ports. So would you remove this as well?
> 
> I personally would: We really don't support Chromium --or boringssl-- any more, right? Besides, the support always was very limited (only a few algorithm combinations -- hence the text you left in). Therefore, I would go even further and would remove as part of this PR Chromium from the list of supported demos -- I did the same with `haproxy` as there was no feedback to the opposite (no-one speaking up voicing interest to use it). Code remained, but no CI any more. That way, it can be resuscitated if someone would pledge the time to maintain it.
> 
> @dstebila @christianpaquin What's your take? We could then even close #189 as there was no activity on this from our side (@Raytonne, please speak up if you see a need for this).

Agreed, unless we get someone speaking up and contributing to maintaining our boringssl and Chromium work.

baentsch:
> Hi - I can help to maintain the boringssl & chromium fork

Great to hear that. Thanks @Raytonne ! I'd then be fine merging this PR as-is.

PR #207, opened 2023-06-12
Revert openvpn base image to debian bullseye

baentsch:
No idea what happened to debian `bookworm` image, but `apt update` simply won't work any more. So switching `oqs-openvpn` docker base image to `bullseye`.


baentsch:
FYI, will merge later today as review seems not needed in order to get CI to pass and to tag release

Issue #111, opened 2021-12-31
Issue With OQS-Chromium


Raytonne:
Hello,

I've already contacted @baentsch regarding this issue, and it might be helpful to post it.

Here is the problem: we've created the following certificate chain:

Common Name | Summary | Public Key Algorithm | Signature Algorithm | Link
:---: | :---: | :---: | :---: | ---
TestCA | Root Certificate | P-384 | ecdsa-with-SHA384 | [view](https://pastebin.com/raw/jAj05vTW)
Test TLS CA | Intermediate Certificate | dilithium2 | ecdsa-with-SHA384 | [view](https://pastebin.com/raw/WETU1aCZ)
dilithium2.webhop.me | Leaf Certificate | dilithium2 | dilithium2 | [view](https://pastebin.com/raw/c1auZzC1)

The root certificate has been imported into Chromium successfully. The website [https://dilithium2.webhop.me (IPv6 Only)](https://dilithium2.webhop.me/) supports `frodo640aes:ntru_hps2048509:bikel1:hqc128:p256_lightsaber:p256_sikep434:p256_sidhp434` key exchange algorithms and is accessible (we checked with OQS-cURL). 

**What We Expect:**
Since the root certificate has been installed and the verification of _root certificate - intermediate certificate_ chain doesn't require any OQS-enabled library, the website should be accessible in Chromium.

**What Actually Happened:**
Chromium returned two errors:

> [10842:10862:1231/225648.339848:ERROR:ssl_client_socket_impl.cc(959)] handshake failed; returned -1, SSL error code 1, net_error -113
>[10842:10862:1231/225648.343227:ERROR:ssl_client_socket_impl.cc(959)] handshake failed; returned -1, SSL error code 1, net_error -207

![Chromium Screenshot 1](https://i.ibb.co/Pg03cHY/OQS2021Q.jpg)

We further found that Chromium failed to verify the _root certificate - intermediate certificate_ chain, but it succeeded in verifying the _intermediate certificate - leaf certificate_ chain.

![Chromium Screenshot 2](https://i.ibb.co/vV9rq1m/OQS2021W.jpg)

We tried to add the intermediate certificate to Chromium; nonetheless, we got the same error as #92.

Any ideas why this happened?

p.s. Wish everyone a joyful New Year! :relaxed:

baentsch:
Thanks for the test and the complete description of your setup! Based on this, two questions:

[...]

PR #210, opened 2023-07-01
Update Chromium build instructions

Raytonne:
* Add instructions to build Chromium on Windows
  - Add QUIC support to the OQS-Chromium (currently quantum safe KEX cannot be used with QUIC on OQS-Chromium)
  - Add quantum safe signature support to QOS-Chromium (solve issue #111 )
* A prebuilt 64-bit Windows Chromium installer can be found [here](https://filebin.net/vjpj3eo13yb3tes2). Note: oqstest_CA is hard coded into the source code of this Chromium

### DIFF
--- a/chromium/README-Linux.md
+++ b/chromium/README-Linux.md
@@ -12,7 +12,7 @@ Note that both cloning the source code as well as building Chromium can take sev
 
 - `git remote add oqs-bssl https://github.com/open-quantum-safe/boringssl`
 - `git fetch oqs-bssl`
-- `git checkout -b oqs-bssl-master oqs-bssl/master`
+- `git checkout -b oqs-bssl-master e2d2587065eacfe97aaae940dd43cd964b71f5b4`
 
 4. In a directory of your choosing, clone and build liboqs as follows:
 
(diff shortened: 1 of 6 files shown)


chromium/README-Linux.md
```diff
@@ -0,0 +1,67 @@
...
+
+0. Navigate to `<CHROMIUM_ROOT>`, and start Chromium by executing `./out/Default/chrome`
+1. Navigate again to the `<CHROMIUM_ROOT>/third_party/boringssl/src` folder, and build OQS-BoringSSL as a standalone project by running `mkdir build && cd build && cmake -GNinja ..`.
+2. Then, in the `build` directory, run `./tool/bssl server -accept 4433 -www -loop -curves <KEX>`, where `<KEX>` can be any key-exchange algorithm named [here](https://github.com/open-quantum-safe/boringssl#supported-algorithms) that is supported by default by Chromium. The [kDefaultGroups array](https://github.com/open-quantum-safe/boringssl/wiki/Implementation-Notes) lists all such algorithms\*.
+3. Load `https://localhost:4433` in Chromium.
+
+An alternative test consists of using the newly built Chromium to access the OQS test server at [https://test.openquantumsafe.org](https://test.openquantumsafe.org) and clicking on any of the algorithm combinations [supported by Chromium](https://github.com/open-quantum-safe/boringssl/blob/master/ssl/t1_lib.cc#L375), e.g., `p256_kyber512`).
```
baentsch, line 53: When looking up this link, it returns "404". The `boringssl` update PR (https://github.com/open-quantum-safe/boringssl/pull/103) also doesn't seem to reference/contain this file.

baentsch: > Do you want me to remove README-Linux.md ? Or we can change the name of README-Linux.md to README-Linux-OUTDATED.md .

No -- if it is meant to contain _current_ instructions how to build for Linux (and you write in the new "README.md" that everything has been tested under Linux), it should be updated with current instructions (containing correct references).

baentsch: > I think we can merge README-Linux and README-Windows into a single file in the future

That would probably be best.

> I also modified the git command so it will fetch a specific boringssl commit

This seems very risky to me: Any future updates to oqs-boringssl will then not be picked up any more. But then again, you agreed to look after that subproject, so I trust you'll update these explicit commit references also here whenever you do changes there.

baentsch:
> Seems it has a problem with BIKE, we'll fix it next time.

OK with me. I'd appreciate a comment in the README stating that though, so that people know what to expect (and not raise a --known-- issue).

baentsch (approved):
LGTM- Thanks for the contribution!

baentsch:
> Thank you! To build the new Chromium we need [boringssl PR 113](https://github.com/open-quantum-safe/boringssl/pull/103), so I think we should merge that PR first. @baentsch

Agreed. Please see review there.

PR #212, opened 2023-07-04
Update ubuntu_x64_provider to large resource class in config.yml

baentsch:
Update suggested & created by CircleCI (as "included in your plan").


dstebila (approved):

dstebila:
Okay.  We should watch to make sure it doesn't push us over some free tier limits, but otherwise, faster is better.

A bunch of line break changes in there, did you mean those?  No worries either way, as long as it doesn't affect the parsing of the YML file.

PR #222, opened 2023-08-07
I(ETF) interop image

baentsch:
Creates an image for interop testing at https://github.com/IETF-Hackathon/pqc-certificates:
- Based on ubuntu
- builds `openssl` (master)
- latest `liboqs`
- `oqsprovider` with all algorithms and [NOPUBKEY_IN_PRIVKEY](https://github.com/open-quantum-safe/oqs-provider/blob/main/CONFIGURE.md#nopubkey_in_privkey) set

Note: Not added to CI: Worthwhile? Current image deposited at https://hub.docker.com/repository/docker/openquantumsafe/oqs-ossl3/general (:interop)


openssl3/Dockerfile-interop
```diff
@@ -0,0 +1,135 @@
...
+# Get all software packages required for builing openssl
+RUN apt install -y build-essential clang libtool make gcc ninja-build cmake libtool wget git
+
+# get current openssl sources
+RUN mkdir /optbuild && cd /optbuild && git clone --depth 1 --branch master https://github.com/openssl/openssl.git
+
+# build OpenSSL3
+WORKDIR /optbuild/openssl
+RUN LDFLAGS="-Wl,-rpath -Wl,${INSTALLDIR_OPENSSL}/lib64" ./config shared --prefix=${INSTALLDIR_OPENSSL} && \
+    make ${MAKE_DEFINES} && make install && if [ -d ${INSTALLDIR_OPENSSL}/lib64 ]; then ln -s ${INSTALLDIR_OPENSSL}/lib64 ${INSTALLDIR_OPENSSL}/lib; fi && if [ -d ${INSTALLDIR_OPENSSL}/lib ]; then ln -s ${INSTALLDIR_OPENSSL}/lib ${INSTALLDIR_OPENSSL}/lib64; fi
```
SWilson4, line 36: Is there an outside chance that both `/opt/openssl32/lib` and `/opt/openssl32/lib64` will exist as directories? If so, might want to edit the logic here so that the symlinks are not created in that case.

openssl3/serverstart-ubuntu.sh
SWilson4, line 1: It seems a shame to have two files which are the same up to choice of shell. What about only using one `sh` script which dynamically launches either `bash` or `sh`?

For example:

`if command -v bash > /dev/null; then bash; else sh; fi`

SWilson4 (approved):
Approved with a couple of (very) minor suggestions.

openssl3/Dockerfile-interop
baentsch: > Is there an outside chance that both /opt/openssl32/lib and /opt/openssl32/lib64 will exist as directories?

No: OpenSSL creates either, not both. I still don't understand the logic which when.

openssl3/serverstart-ubuntu.sh
baentsch: > It seems a shame to have two files which are the same 

Up to here I agree...

> up to choice of shell. 

... not here, though: The true difference is the env var presence check that differs between `sh` and `bash`. If you'd know of a way to "unify" that, I'd be really grateful.

SWilson4: I believe that `sh` also supports the `-z` operator, so checking to see if `KEM_ALG` is set can be done in the same way for both.

Taking a closer look, I don't think the `==` operator is specified by the `sh` standard, so lines 5-7 in the `sh` script might not actually be doing anything at the moment. (`sh` complains about an unexpected operator on my local machine, for instance.)

baentsch: > I believe that sh also supports the -z operator

It does: Thanks! Will streamline things then.

baentsch: @SWilson4 If you'd like to take another look after the latest commit, please do. I'd merge otherwise if CI passes.

SWilson4: Looks good to me!

PR #223, opened 2023-08-08
Update test.openquantumsafe.org & resolve some issues

bhess:
Updates the build for test.openquantumsafe.org:
- Updates oqs-provider to 0.5.1

Fixes a few issues:
- ssl groups for OQS-kems were not correctly resolved to strings. 

Example:
```
/curl -v --cacert cert.pem --curves x25519_kyber512:kyber768 https://test.openquantumsafe.org:6245/
...
Client-side KEM algorithm(s) indicated:
0x2f39:0x023c
```

The update resolves them correctly:
```
/curl -v --cacert cert.pem --curves x25519_kyber512:kyber768 https://test.openquantumsafe.org:6245/
...
Client-side KEM algorithm(s) indicated:
x25519_kyber512:kyber768
```
-> needed to patch nginx to use `SSL_group_to_name` instead `OBJ_nid2sn` when using OpenSSL >= 3.0.

- signature/kem response was not available with the ports that offer the default groups. Example: 
https://test.openquantumsafe.org:6043 returned `Successfully connected using !`. With the update it returns `Successfully connected using rsa3072-x25519!`

-> updated interop.conf to not hardcode the negotiated group in the response html but instead use nginx's `$ssl_curve` variable available from v1.21.5. Also needed to patch nginx to correctly resolve the string when using OpenSSL >= 3.0.

- server html response was malformed (missing `</html>` tag)

The update is already applied to https://test.openquantumsafe.org.
May see if there is interest in nginx upstream to apply the patch there.

### DIFF
--- a/nginx/fulltest-provider/Dockerfile
+++ b/nginx/fulltest-provider/Dockerfile
@@ -7,7 +7,7 @@ ARG LIBOQS_VERSION=0.8.0
 
 ARG OPENSSL_VERSION=master
 
-ARG OQS_PROVIDER_VERSION=0.5.0
+ARG OQS_PROVIDER_VERSION=0.5.1
 
 ARG LIBOQS_BUILD_DEFINES="-DOQS_DIST_BUILD=ON"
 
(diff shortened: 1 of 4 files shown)


baentsch (approved):
LGTM. Thanks for the improvements.

PR #233, opened 2023-10-11
Update Chromium Linux Build Instructions

Raytonne:


baentsch:
> We'll test the Linux instruction tomorrow. Let's hold this PR until next week

Thanks for the fast reaction -- and foreseeing my concern: Publishing something without having tested it IMO is not good practice.

Please also consider automating more: From what I saw you're asking users to add quite a bit of code that can (most likely _will_) be outdated pretty quickly (algorithm names and OIDs, for example). That's why we so far opted for a patch-based mechanism: That way it becomes visible automatically if something (triggered by up- and downstream changes) goes wrong and we don't lead users astray, i.e., have users "discover" when something breaks (I know this is pretty much "standard procedure" for commercial software but that does not mean that open source software needs to follow this approach :).

PR #237, opened 2023-10-18
curl version change to work around bug

### LINKED ISSUE #236: curl --write-out time_appconnect for SSL/TLS Handshake variable is faulty
`--write-out` or  `-w`, outputs text and information after a transfer is completed. It offers a large range of variables that you can include in the output, variables that have been set with values and information from the transfer. 
According to [curl manuals](https://everything.curl.dev/usingcurl/verbose/writeout) `time_appconnect `variable for the TLS/SSL Handshake time.
So order should be `time_namelookup` < `time_connect` < `time_appconnect `< `time_pretransfer` < `time_starttransfer` < `time_total `
First time I noticed this problem was when i tried to analyze the times for the [OQS Testserver](https://test.openquantumsafe.org/).
For this command:

`docker run -v ``pwd``:/ca -it openquantumsafe/curl curl --cacert /ca/CA.crt https://test.openquantumsafe.org:6001 --curves bikel1 -s -w "%{time_namelookup},%{time_connect},%{time_appconnect},%{time_pretransfer},%{time_starttransfer},%{time_total}" -I`

I get this output:

`0.038224,0.238271,0.517486,0.279321,0.311949,0.312131`

First I thought this might be some faulty implementation of the server, but i have the same issue with other websites. The normal curl version works fine.
If I missed something feel free to comment.

baentsch:
Fixes  #236

### DIFF
--- a/curl/Dockerfile
+++ b/curl/Dockerfile
@@ -1,7 +1,7 @@
 # Multi-stage build: First the full builder image:
 
 # define the Curl version to be baked in
-ARG CURL_VERSION=7.87.0
+ARG CURL_VERSION=7.81.0
 
 # Default location where all binaries wind up:
 ARG INSTALLDIR=/opt/oqssa


baentsch:
Prodding @dstebila @bhess for review & approval: After the [oqs-provider release](https://github.com/open-quantum-safe/oqs-provider/releases/tag/0.5.2) this PR can close the one known defect on oqs-demos prior to tagging it and its docker images "0.9.0" and considering the release round complete.

baentsch:
Oops -- one more thought: Didn't we want to update/release `openssh`, too, before tagging `oqs-demos`? @praveksharma ? But then again, that project seems dead: No activity for about 6 months; also I didn't find an option to tell (algorithm) versions apart in different builds, so why bother? Or is there a way to see, e.g., which McEliece/liboqs version is used in a specific `oqs-openssh` version? If not, there really may be no reason to release `oqs-openssh`: The most current version will simply run with `liboqs` 0.9.0 code and accordingly not interoperate with 0.8.0 builds -- but that doesn't seem to be a concern anyway, right @dstebila ?

baentsch:
Update to the above: When https://github.com/open-quantum-safe/openssh/pull/148 gets merged (not by me but someone knowledgeable in OpenSSH), tagging a release would be a minor thing: Would you want to do that too, @praveksharma ? After that, merging this PR is really the last thing: When done, all resultant automatically generated docker images should be tagged 0.9.0 and we can consider the 0.9.0 release cycle complete.

dstebila (approved):

PR #241, opened 2023-10-25
nginx server: update to oqs-provider 0.5.2 & config update

bhess:
Updates Dockerfile to generate package for test.openquantumsafe.org:
- update oqs-provider to 0.5.2 tag
- config update: sets `keepalive_timeout 0;`: to avoid unnecessary keep-alives and close connection immediately after request is fulfilled.

Update is already deployed at test.openquantumsafe.org, fulltest.sh passes.


baentsch (approved):
Thx!

PR #245, opened 2023-11-02
add Chromium usage instructions [skip ci]

baentsch:
As per https://github.com/open-quantum-safe/oqs-demos/discussions/243#discussioncomment-7458152. Feedback welcome @pi-314159 .


pi-314159:
LGTM. A small suggestion:

> Execute `./chrome` in the directory...

      Execute `./chrome` or `chrome.exe` in the directory...

When we edit README-Linux and README-Windows next time, we'll add notes about how to create a installer on Windows and how to pack the build directory on Linux.

baentsch:
@bhess, this contains an update that may be worth while deploying to the test server.

dstebila (approved):

bhess (commented):
Thanks @baentsch. 
However, for the test server update, the following page would need to be updated:
https://github.com/open-quantum-safe/oqs-demos/blob/main/nginx/fulltest-provider/index-template.

(The version in the [fulltest](https://github.com/open-quantum-safe/oqs-demos/blob/main/nginx/fulltest/index-template) folder is deprecated / used to build against the old OpenSSL 1.1.1 fork)

baentsch:
> The version in the [fulltest](https://github.com/open-quantum-safe/oqs-demos/blob/main/nginx/fulltest/index-template) folder is deprecated / used to build against the old OpenSSL 1.1.1 fork

Thanks for this correction. Now fixed. What about removing the old directory then?

bhess:
Ok with me to delete the old folder.

baentsch:
> Ok with me to delete the old folder.

Looking at this from a code-origin traceability perspective deleting `fulltest-provider` seems more correct: Would you want to do such PR, @bhess such as to have your code changes properly attributed to you?

bhess:
Added a PR #249 for deleting the old directory and renaming fulltest-provider.

I also deployed the update from this PR at test.openquantumsafe.org. A test run works fine.
However @baentsch, I noticed that the "0.9.0" tag is missing on dockerhub for [openquantumsafe/curl](https://hub.docker.com/r/openquantumsafe/curl/tags), which is used for the test run. Could you please publish this tag? The ":latest" tag isn't compatible anymore because of the recent HQC update in liboqs.

baentsch:
Oops -- looks like this question slipped my attention. It should now be addressed as/when https://github.com/open-quantum-safe/oqs-demos/issues/257 gets resolved. The point itself is moot by now given we are about to completely eliminate references to 0.9.0/0.9.1.

PR #258, opened 2024-01-17
Fix broken relative link in nginx/fulltest/README.md

aes512:
The doc points at `build-server.sh` when the filename  is actually `build_ubuntu.sh`.


baentsch (approved):
Thanks for the fix!

PR #263, opened 2024-01-24
convenience script to release specific OQS component versions [skip ci]

baentsch:
Script to be probably used rarely creating and publishing docker images for specific `liboqs` and `oqsprovider` versions. Tested to release "0.9.2" versions of all supported docker images.

### DIFF
--- a/scripts/buildnpush.sh
+++ b/scripts/buildnpush.sh
@@ -1,0 +1,31 @@
+#!/bin/bash
+
+# Helper script to build supported docker images for a specific liboqs/oqsprovider tag/branch and push to docker hub
+# It assumes the user has already successfully logged in to the docker hub namespace used
+
+# set to ascertain docker build cache is not used
+#export NOCACHE="--no-cache"
+
+# docker hub name space to push images to
+export OQS_REPO=openquantumsafe
+
+# docker image tag to set
+export RELEASE_TAG=0.9.2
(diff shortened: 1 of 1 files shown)


scripts/buildnpush.sh
```diff
@@ -0,0 +1,31 @@
...
+# It assumes the user has already successfully logged in to the docker hub namespace used
+
+# set to ascertain docker build cache is not used
+#export NOCACHE="--no-cache"
+
+# docker hub name space to push images to
+export OQS_REPO=openquantumsafe
+
+# docker image tag to set
+export RELEASE_TAG=0.9.2
```
SWilson4, line 13: I hate to prolong the review process for a script that we may not use again. But... would it be better if these strings were passed as command line arguments, and the script does nothing if all three are not explicitly set?

SWilson4 (approved):
This looks good to me, though I haven't tested it myself. I left one suggestion that you may or may not want to put time into implementing @baentsch.

scripts/buildnpush.sh
baentsch: > for a script that we may not use again

That very thought drove my "motivation" to (not) do this in a more parameterized manner. Let's leave this to interested (potential/new) contributors...

dstebila (approved):

PR #275, opened 2024-05-13
reduce further and document better support level of single integrations

baentsch:
Follow-up to https://github.com/open-quantum-safe/tsc/issues/2

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@ As the level of interest in providing and maintaining these integrations for pub
 
 We are explicitly soliciting contributors to maintain those integrations labelled "unsupported".
 
-Currently supported packages:
+Currently available integrations at their respective support level:
 
 |                  | **Build instructions**                 | **Pre-built Docker image or binary files**                                                                                   | Support? |
 | ---------------- | -------------------------------------- | ---------------------------------------------------------------------------------------------------------------------------- | -------- |
@@ -47,8 +47,6 @@ All modifications to this repository are released under the same terms as [liboq
 
 ## Team
 
-The Open Quantum Safe project has been taken over by LinuxFoundation within [PQCA](https://pqca.org). 
-
 ## Contributors to oqs-demos include:
 
     Christian Paquin (Microsoft Research)
@@ -69,4 +67,5 @@ The Open Quantum Safe project has been taken over by LinuxFoundation within [PQC
 
 Most effort in this project has been provided by individual contributors working in their own time and out of personal interest to see how PQ crypto integrates into existing software stacks.
 
-Basic financial support for the development of Open Quantum Safe has been provided by various institutions such as Amazon Web Services and the Canadian Centre for Cyber Security. Also, several research institutions and companies have dedicated programmer time to contribute source code to OQS, including but not limited to University of Waterloo, Amazon Web Services, evolutionQ, Microsoft Research, Cisco Systems, wolfSSL, University of Mauritius, IBM Research, sandboxAQ and Fachhochschule Nordwestschweiz. Research projects which developed specific components of OQS have been supported by various research grants.
+This project is part of [Open Quantum Safe](https://openquantumsafe.org/news/).
+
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -66,8 +67,6 @@ The Open Quantum Safe project is led by [Douglas Stebila](https://www.douglas.st
 
 ## Acknowledgments
 
-Financial support for the development of Open Quantum Safe has been provided by Amazon Web Services and the Canadian Centre for Cyber Security.
-
-We'd like to make a special acknowledgement to the companies who have dedicated programmer time to contribute source code to OQS, including Amazon Web Services, evolutionQ, Microsoft Research, Cisco Systems, IBM Research and Fachhochschule Nordwestschweiz.
+Most effort in this project has been provided by individual contributors working in their own time and out of personal interest to see how PQ crypto integrates into existing software stacks.
```
dstebila, line 70: If you would like the README to include a statement acknowledging individual contributors, that's fine. However this statement goes beyond that in ways that I do not think should appear in the README.

baentsch: In which way? How would you rephrase?

dstebila: How about:
```suggestion
This project has benefited substantially from the efforts of individual volunteer contributors.
```

README.md
```diff
@@ -46,7 +47,7 @@ All modifications to this repository are released under the same terms as [liboq
 
 ## Team
 
-The Open Quantum Safe project is led by [Douglas Stebila](https://www.douglas.stebila.ca/research/) and [Michele Mosca](http://faculty.iqc.uwaterloo.ca/mmosca/)at the University of Waterloo.
+The Open Quantum Safe project has been taken over by LinuxFoundation within [PQCA](https://pqca.org).
```
dstebila, line 50:
```suggestion
The Open Quantum Safe project is part of the Linux Foundation within [PQCA](https://pqca.org). 
```

baentsch: This change would completely eliminate the history of OQS as a free/"meritocratic", non-LF project. If you don't like the term "taken over" what about "After 7 years of stewardship by UWaterloo, control of the project changed to the newly established PQCA, a Linux Foundation alliance."?

README.md
baentsch: This text proposal is a true statement about practically any OSS project -- and it doesn't mention the "full software stack" aspect that makes/made this work particularly challenging; your wording could equally well be applied to all OQS sub projects or be left away entirely.

And for the avoidance of doubt: I do not want to stress the aspect of volunteerism (I presume meaning people working for free/outside of employment) but of individuals going beyond their assigned tasks; this for example also includes people like @xvzcf and practically all external contributors -- of whom I don't know their employment status -- but do know about (and wanted to acknowledge with my proposal) their personal interest and commitment.

dstebila:
```suggestion
This project has benefited substantially from the efforts of individual contributors motivated to see how PQ crypto integrates into existing software stacks.
```

README.md
dstebila: This is documented to my satisfaction at https://openquantumsafe.org/news/, so I don't have the need to repeat it in each README.  But if we do want to include it in each README, then I'd just copy the text from the website.

baentsch: > then I'd just copy the text from the website

Sounds a bit 20th century :-) What about instead of the text simply linking (actually from/for all sub project's README's) to there? This may actually remove lots of redundant text (e.g., regarding funding) in many sub projects.

dstebila: Sure, we can do that.

README.md
baentsch: I still see a marked difference between the wording "benefited substantially from" and "most effort has been provided" (by individual contributors): Here's the count of PRs done by "individual contributors" and contributors doing this as part of their employment:

145 to 36 contributions -- provided by 16 individuals to 4 employed ones. Thus, in both counts about 80% has been done by individual contributors, in my eyes justifying the wording I proposed.

dstebila: Resolve this as you see fit.

baentsch: Done

README.md
baentsch: Added link to OQS news.

PR #321, opened 2024-11-20
Move CI to GitHub Actions

### LINKED ISSUE #213: Create cross-platform docker images in github
- Become independent from CircleCI for docker image creation
- Create multi-architecture images
- Publish those to [Github Packages](https://github.com/features/packages) and [Docker Hub](https://hub.docker.com/)
- Automate release tagging for docker images

### LINKED ISSUE #284: Automate and streamline docker image generation
Upon merge to "main" or when triggered by upstream build/release events, this project generates ready-made docker images at docker hub for users interested in "just using" OQS.

Currently, manual intervention right at the moment of completion of downstream CI is necessary to tag a specific docker image corresponding to a specific oqs-provider/liboqs release when the docker image has been created by CircleCI. If this moment is missed, the specific version needs to be manually created with the specific desired oqsprovider/liboqs version manually inserted to the Dockerfile(s) as subsequent CI runs overwrite the "latest" docker image. If using Github CI, a simple re-labeling of the correct commit is required as this commit ID is stored at docker hub (right now only for "openquantumsafe/openssl3").

This issue is to suggest streamlining and fully automating this:
- Generate docker images for [all still supported OQS integrations](https://github.com/open-quantum-safe/oqs-demos/pull/275) via GH CI, [moving away from CCI](https://github.com/open-quantum-safe/oqs-demos/issues/213)
- Upon an upstream GH release event, automatically tag the generated docker images with the upstream version ID

Tagging @ryjones fyi as I learned from @dstebila that you seem to know "everything github" and this issue may be straightforward for you to implement.

### LINKED ISSUE #294: Switch off CircleCI
The project already started to move to GH CI. This issue is to suggest completing the move off CCI.

ajbozarth:
PR for moving our CI from CircleCI to GitHub Actions. This work includes decoupling build and test steps from the push step as well as support multi-platform builds.

Implementation notes:

- Each demo has it's own workflow
- A workflow will only run when code specific to that demo is changed
- ghcr.io is used for build and test steps to remove reliance on DockerHub credentials in forks
    - DockerHub auth will occur only when push is run
- Build and test steps can also run against liboqs and oqs-provider main branches by manually triggering a workflow and setting the `build_main` flag to true.
- A Run All workflow is included that triggers all the other workflows, this will be useful for times when all the CI need to be manually triggered, such a for a release or to test against lliboqs main.
    - I will be adding workflows to liboqs and oqs-provider to trigger this as followup work
- The `latest` image will be pushed to DockerHub and ghcr.io when a workflow is run on the main branch, when not a fork, not a PR, and not building against liboqs main branch
    - The `latest` image is actually a manifest that points to new platform-based images `latest-x86_64` and `latest-arm64`
    - If a build is manually triggered, the input value `release_tag` can be set that will replace `latest` in the above naming convention.
- linux.yml has been deleted and quic.yml has been update to follow the above push rules, but otherwise current workflows have not been changed

Fixes #294 Fixes #213 Fixes #284


ajbozarth:
Note that I have only opened this draft to allow for iterative feedback. It is far from complete and the current workflow is just my first working example (or MVP for agile folks)

ajbozarth:
Final update before I go on vacation til Monday. I added workflows for curl and httpd and moved the build against liboqs main from a matrix build to a trigger option. I've updated the description and previous commits to clarify my latest changes.

ajbozarth:
> I would like to add a linter for Docker files, such as [Hadolint](https://github.com/hadolint/hadolint) or should we rely on [MegaLinter](https://github.com/oxsecurity/megalinter) instead? What do you think?

This is a great idea, I'm not sure on which linter to use, I've never linted Dockerfiles before. Creating a linter GitHub Action would be a great task you can do in parallel to this PR. You would most likely need to do a lot of small updates to get the linter to pass initially though.

> I have noticed that some GitHub Actions are not using the latest versions

How so? The actions I added check out and run the latest code locally, there's currently no versioning at all.

> Wouldn’t it be better to rely on -j$(nproc) instead

This is something I'm aware of, on CircleCI it was being hard-coded to `-j 18` and had a comment on the number of available cores in circle CI being 35. I tried to set it to `-j$(nproc)` in my workflows but it cause a string escape issue so I instead set it to `-j4` since GitHub Actions only has 4 available cores. I plan to go back and see if I can get `-j$(nproc)` to work, but I didn't want that small issue to block my overall progress too long, so I set it aside for now.

ajbozarth:
A note here on my most recent update:

I had previously enabled triggering each workflow with a flag to use liboqs/oqsprovider main branches, I have moved this to a callable action (meaning it can be triggered by another workflow but not manually triggered). Instead I have created `build.yml` which is both trigger able and callable and runs all the workflows with the given input flag. This will enable the most common use case: letting liboqs or oqsprovider to remotely trigger a build and test of the demos using the latest code in their own workflows.

ajbozarth:
I just pushed a few more workflows, included in them are locust and wireshark, both of which are just build and have no test step:

- In the case of wireshark this was how it was already in circleCI since it's a UI demo and doesn't have a way to run without X11. @Hayyaaf if you disagree with this and have an idea on a test command feel free to chime in as the new maintainer, and I can add it.

- For locust though I am unsure if we want to include a test step, I read through the README and USAGE files and all the "check your work" instructions seem to rely on a UI. @davidgca as the maintainer of locust is there a command or set of commands you recommend using for the test step? Feel free to look at the other workflows I added in this PR for examples of how we test other demos. Including CI tests is left to your discretion as the maintainer.

ajbozarth:
After playing with the CI a bit more I believe I have an implementation for push, I did check if it pushed to my personal ghcr.io successfully and it did. You can see my first try in the openssl3 workflow. It includes pushing to docker hub and ghcr.io whenever the repo is open-quantum-safe/oqs-demos (ie not a fork) and the branch is main. This prevents pushes when running CI in forks or PRs. I also included a new user input when triggering the workflow to set the tag version to push as I pitched in my previous comment.

Further issues/discussion based on my first bullet [above](https://github.com/open-quantum-safe/oqs-demos/pull/321#issuecomment-2499246961):
- Does this actually work for DockerHub? I can't test this using my own credentials due to IBM's policies on DockerHub.
- Should the workflow add more "conditions" to when it pushes? Such as "don't push if building liboqs from main" or adding an input flag and "only push if the release flag is set" to allow running of the workflows manually on main for testing without a push.

Also I looked into the other workflows and have a few more insights:
- `docker-scan.yml` the scanning workflow this uses require authentication with DockerHub, so there's no way to run it on PRs, thus it should be left alone
- `quic.yml` while reading it I noticed it does pushes with a cron job, I thought that was odd and realized that the QUIC Dockerfiles do not pin their versions, so that workflow regularly rebuilds the QUIC demos using the latest liboqs and pushes the images. If we're ok with the QUIC demos' behaving different from the rest I say we leave this alone, though it may be worth adding some conditionals so the build and test steps can run on PRs (I may just do this and push it if I have a moment)

lastly it seems the wireshark demo runs a oqsprovider test script during build, which fails when run on the CI (but passes locally). @Hayyaaf what is the reason for running that test script in the dockerfile? Is it something we can remove? otherwise we'll need to debug why it causes the build to fail in the CI but not locally.

davidgca:
> * For locust though I am unsure if we want to include a test step, I read through the README and USAGE files and all the "check your work" instructions seem to rely on a UI. @davidgca as the maintainer of locust is there a command or set of commands you recommend using for the test step? Feel free to look at the other workflows I added in this PR for examples of how we test other demos. Including CI tests is left to your discretion as the maintainer.

Hi @ajbozarth ,

I think we can integrate a test steps without modifying the locustfile.py and locust compose. Locust has an non-webUI execution as its describe [here](https://docs.locust.io/en/stable/running-without-web-ui.html#running-without-the-web-ui) similar to  for example ab, but need some specific modification in our code. Here's how you could run the tests in 4 steps, is not the most elegant solution but maybe is helpful for this purpose:
1. Launch the Locust environment with parameters: Use Docker Compose to set up the Locust master and worker, specifying the quantum-safe group and target host:

   `LOGGER_LEVEL=DEBUG HOST=https://localhost:4433 GROUP=kyber768 docker compose up --scale worker=1`

2. Start the quantum-safe server: Use docker exec to launch the server with OpenSSL's OQS support directly inside the worker container:
`
docker exec -ti locust-worker-1 openssl s_server -cert bin/CA.crt -key bin/CA.key -www -tls1_3 -groups kyber768`

3. Run the test for 30 seconds: Trigger the Locust master to execute the test by sending a POST request to its API:
```
curl -X POST -H "Content-Type: application/json" \
        -d '{"user_count": 10, "spawn_rate": 2, "run_time": "1m", "host": "https://localhost:4433"}' \
        http://locust-master:8089/swarm
```
4. Download results in csv format:
`curl 'http://localhost:8189/stats/requests/csv' -o results.out `
with this output:
```Type,Name,Request Count,Failure Count,Median Response Time,Average Response Time,Min Response Time,Max Response Time,Average Content Size,Requests/s,Failures/s,50%,66%,75%,80%,90%,95%,98%,99%,99.9%,99.99%,100%
GET,Group kyber768 /customers,8,0,11,16.75,10,27,0.0,0.2671710359431904,0.0,13,25,26,26,27,27,27,27,27,27,27
GET,Group kyber768 /devices,12,0,14,14.833333333333334,9,26,0.0,0.4007565539147856,0.0,14,14,18,18,25,26,26,26,26,26,26
,Aggregated,20,0,13,15.6,9,27,0.0,0.667927589857976,0.0,14,15,25,25,26,27,27,27,27,27,27
```
we maybe take the last line that give us the summary of the test: number of requests and number of errors

This setup validates the integration of quantum-safe algorithms in TLS 1.3 and tests the system under load without requiring manual interaction or modifications to the existing Locust test scripts.

I'm not familiar with github actions but maybe the steps should be:

```
name: Locust Quantum-Safe Test

on:
  push:
    branches:
      - main

jobs:
  locust-test:
    runs-on: ubuntu-latest

    steps:
    - name: Checkout repository
      uses: actions/checkout@v3
      # Ensure the repository is available for Docker Compose and scripts.

    - name: Set up Docker
      uses: docker/setup-buildx-action@v2
      # Prepare Docker for the Locust and OpenSSL environment.

    - name: Launch Locust in Docker Compose
      run: |
        LOGGER_LEVEL=DEBUG HOST=https://localhost:4433 GROUP=kyber768 docker compose up -d --scale worker=1
      # Launch the Locust environment with a single worker for this test case.

    - name: Start OpenSSL server
      run: |
        docker exec -d locust-worker-1 openssl s_server \
          -cert bin/CA.crt \
          -key bin/CA.key \
          -www \
          -tls1_3 \
          -groups kyber768
      # Start the OpenSSL quantum-safe server inside the Locust worker container.

    - name: Run Locust load test
      run: |
        curl -X POST -H "Content-Type: application/json" \
          -d '{"user_count": 10, "spawn_rate": 2, "run_time": "1m", "host": "https://localhost:4433"}' \
          http://localhost:8089/swarm
      # Trigger the Locust load test by sending a POST request to the Locust master API.

    - name: Wait for test to finish
      run: sleep 70
      # Allow the test to complete by waiting longer than the specified run-time.

    - name: Download JSON report
      run: |
       curl 'http://localhost:8189/stats/requests/csv'  -o locust_report.out
      # Fetch the Locust test results in CSV format using the /stats/requests API.

    - name: Upload JSON report
      uses: actions/upload-artifact@v3
      with:
        name: locust-report
        path: locust_report.out
      # Save the JSON report as an artifact so it can be downloaded after the workflow completes.

    - name: Tear down Docker Compose
      run: docker compose down
      # Stop and remove all Locust-related containers to clean up the environment.

```

Let me know if you'd like any further help

Cheers,
David

ajbozarth:
@davidgca I tried out your suggested steps both locally and on the CI and the curl call just won't work, I tried playing with it, switching the port from 8089 to 8189 change the error from `Couldn't connect to server` to `cannot access local variable 'user_count'`. I also did a bit of googling and afaik locust's API won't take call from curl. I did confirm that if we can run that step then the rest will work by just opening the webUI and running the load test there. 

If you can figure out a way to get that curl call to work then I'll add the test, otherwise let's leave it without a test for now.

While testing your demo though I did notice that it hard coded the docker tag in the docker compose, this can cause issues with users who checkout the `latest` tag, is hard-coding that tag necessary for docker compose? I have minimal experience with docker compose.

davidgca:
> @davidgca I tried out your suggested steps both locally and on the CI and the curl call just won't work, I tried playing with it, switching the port from 8089 to 8189 change the error from `Couldn't connect to server` to `cannot access local variable 'user_count'`. I also did a bit of googling and afaik locust's API won't take call from curl. I did confirm that if we can run that step then the rest will work by just opening the webUI and running the load test there.
> 
> If you can figure out a way to get that curl call to work then I'll add the test, otherwise let's leave it without a test for now.
> 
> While testing your demo though I did notice that it hard coded the docker tag in the docker compose, this can cause issues with users who checkout the `latest` tag, is hard-coding that tag necessary for docker compose? I have minimal experience with docker compose.

Sorry ,my mistake, I pasted a previous curl attempt. The first curl that should start the locust cluster is:
`curl -X POST -H 'content-type: application/x-www-form-urlencoded' --data-raw 'user_count=1&spawn_rate=1&host=https%3A%2F%2Flocalhost%3A4433&run_time=30s' http://localhost:8189/swarm`

the second curl, for download the test after 30 seconds is the same as I suggested before:
`curl 'http://localhost:8189/stats/requests/csv' -o results.out`
let me know if now these steps are working to you.

ajbozarth:
@davidgca thanks for the help, that got it working. I just pushed the updated workflow based on your suggested code

ajbozarth:
I just pushed a small fix to the locust workflow, while debugging it I found that all the arm64 builds throw warnings during their tests when the docker images are run. `The requested image's platform (linux/arm64) does not match the detected host platform (linux/amd64/v3) and no specific platform was requested` I'll need to add a `--platform ${{ matrix.platform }}` to every `docker run` in all the workflows. Making a note of it here since I may not get to it before Thanksgiving

Hayyaaf:
@ajbozarth Thanks for sharing this update. Could you guide me through the process of contributing to your PR? I am facing some difficulties and would appreciate your guidance.

ajbozarth:
>  Could you guide me through the process of contributing to your PR? 

In general in open source, a PR has one contributor and many reviewers. So as a reviewer you can leave comments and code suggestions, then I as the PR author address them. There are exceptions to this where multiple contributors work on a single PR together, but this is less common in open source.

It's also worth noting this is one of the reasons PR's typically should only handle a single feature at a time, so that other contributors can still open other PRs working on other features in parallel without blocking each other.

In the case of the GitHub Actions work this separation is done with separate workflows, so while any work extending the CI workflows I'm creating here would be blocked by my PR, you could for instance create separate workflows and contribute them in a different PR in parallel. For example, as we discussed previously, you could create a `lint.yml` workflow for linting similar to how the `docker-scan.yml` does a code scan.

@Hayyaaf I hope that helps answer you question, sorry if I rambled a bit, I know your new to open source and just was to help :)

Hayyaaf:
@ajbozarth Thanks Alex, I really appreciate your help and guidance!

Regarding the lint.yml, in the next couple of hours, I’ll provide you with the code. Since there are many Dockerfiles across different folders, I’ll try to figure out a way to make it work efficiently for all of them. Does that sound good to you?

baentsch:
Looking at this PR can I ask whether it would be possible to create a script or something similar that generates most of what seems like "boilerplate code"? Or is there a concept for code sharing (like #include or macros in C) for GH CI workflows? That way, a review would become much easier and less failure-prone to overlook something (in "just the same code") . Also, future maintenance (and adding of new integrations) would become much simpler.

.github/workflows/locust.yml
```diff
@@ -55,4 +55,45 @@ jobs:
...
+      - name: Wait for test to finish
+        run: sleep 70
+
+      # Fetch the Locust test results in CSV format using the /stats/requests API.
+      - name: Download and Check JSON report
+        working-directory: ./locust
+        run: |
+          curl 'http://localhost:8189/stats/requests/csv' -o results.out &&
+          cat results.out &&
+          grep kyber768 results.out
```
davidgca, line 99: I think is better to grep "Aggregated" line. This line is the summary (sum, mean depends of the statistics) for all previous lines

ajbozarth: I included the `cat` command for displaying the output, the `grep` is meant to be the "test" checking that the PQ algorithm was used as expected, thus why I greped for kyber768.

ajbozarth:
@baentsch here's a long answer to your comment, I can explain better on Tuesdays call if you are still worried after reading. I'm open to switching up my strategy if it will be better overall for the project.

> Looking at this PR can I ask whether it would be possible to create a script or something similar that generates most of what seems like "boilerplate code"?

I believe there is a way to write a template for creating new test cases, but afaik there's no way to dynamically create workflows with scripts

> Or is there a concept for code sharing (like #include or macros in C) for GH CI workflows?

This concept does exist for `job` steps, in fact the `uses` in a step is just calling another workflow (such as the various docker ones). Sadly afaik there is no such option for the `on` and `env` sections, those must be fully defined in each workflow file. (this is actually where a lot of my "copy-paste" code is.

> That way, a review would become much easier and less failure-prone to overlook something (in "just the same code") . Also, future maintenance (and adding of new integrations) would become much simpler.

I spent a lot of time researching ways to better automate and remove duplication from the workflows, but the largest hangup I hit was the fact that this is a monorepo, which has very little support in GitHub Actions. (ie lots of separate builds inside subdirectories).

I did experiment with combining all the tests into a large single workflow file with each demo having it's own job, but that only removed the duplicate `on` code and actually removed one of the best features: the fact that each demo's workflow will actually only run when edits are made to it's subdirectory. It was also impossible to leverage a metric to build or push the demos since the demos don't all have the same build commands. 

Lastly, as for trying to simplify duplication in the job steps, I've already done a lot of that, most of the steps in each workflow leverage the `use` feature and call other workflows with args.

---

Personal Opinion: Though there is a lot of duplicate "code" in these workflows initially, by having them separate it allows future changes to any specific demo to edit it's build/test/push steps to met it's own unique situation.

baentsch (changes requested):
Thanks for this work @ajbozarth . Please see the two single comments (the latter of course applying to all jobs). I'd think what would be needed is some documentation in CONTRIBUTING.md (or somewhere you deem suitable) as to what this now demands  of contributors: It seems everyone now has to have a ghcr.io account, for example (?). Also there (or somewhere else) would be good to add an explanation as to how these runs can be manually triggered. And a question to aid my understanding: Did I get it right that manual triggers don't push? Or can the push action also be triggered upon a "manual start" (how?)? But again, all that would be good in some documentation (maybe just a link collection to existing documentation explaining how everything hangs together). Lastly, I saw some inefficiencies that may be worth while removing for "main" branch test runs, like installation of openssl documentation that wouldn't be necessary for those tests -- but that's a future issue (maybe worth while a separate issue to not lose track of once this lands). Another thought looking forward: Would it make sense to already now switch over all "Kyberxxx" based tests to MLKEM or another KEM to avoid having to redo this when Kyber is removed in the very close future? Final question: Did all usage tests that are currently implemented in CircleCI get moved to this PR or should I manually check this?

.github/workflows/openssl3.yml
```diff
@@ -0,0 +1,91 @@
...
+      - uses: docker/setup-buildx-action@v3
+      - uses: docker/login-action@v3
+        if: env.push == 'true'
+        with:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
+      - uses: docker/login-action@v3
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
```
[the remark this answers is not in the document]
ajbozarth: It is correct that every contributor has ghcr.io credentials in so much that ghcr.io credentials are GitHub credentials, so as long as you're logged into GitHub your ghcr.io credentials are available in this env var.

ajbozarth:
> I'd think what would be needed is some documentation

I agree, I will add documentation

>  It seems everyone now has to have a ghcr.io account

As I mentioned in my inline comment, a GitHub account is a ghcr.io account, so everyone has one

> be good to add an explanation as to how these runs can be manually triggered

I will make sure to include this in the docs

> Did I get it right that manual triggers don't push? Or can the push action also be triggered upon a "manual start" (how?)?

In my current implementation seen in openssl3.yml any workflow run, automatic or manually triggered, will push as long as it's being run on the main branch of the original (non-fork) repo. I have an open question on this that I will add as a separate inline comment.

> Lastly, I saw some inefficiencies that may be worth while removing for "main" branch test runs, like installation of openssl documentation that wouldn't be necessary for those tests -- but that's a future issue (maybe worth while a separate issue to not lose track of once this lands). Another thought looking forward: Would it make sense to already now switch over all "Kyberxxx" based tests to MLKEM or another KEM to avoid having to redo this when Kyber is removed in the very close future?

I agree these are good follow up issues for separate PRs

> Final question: Did all usage tests that are currently implemented in CircleCI get moved to this PR or should I manually check this?

All the tests in CircleCi were moved over "as is" to these workflows except the QUIC ones. Given QUIC has recently been reworked and already had a GitHub workflow, I ignored those circleCI tests.

## Next Steps

Based on the above comments I will start work on the following:

- [ ] Adding the push step to each workflow
- [ ] Update the build all workflow to support the release_tag input
- [ ] Delete the build.yml and circleci config.yml files
- [ ] Add documentation
- [ ] Update the `env.push` value based on the inline discussion I'll be adding below

Note: I'll be out tomorrow through the weekend, so I will get this done early next week (might do some of it today)

.github/workflows/openssl3.yml
```diff
@@ -0,0 +1,91 @@
...
+      release_tag:
+        description: "Which docker tag to push to"
+        required: false
+        type: string
+
+env:
+  build-args: |
+    LIBOQS_TAG=main
+    OQSPROVIDER_TAG=main
+  push: ${{ github.repository == 'open-quantum-safe/oqs-demos' && github.ref == 'refs/heads/main' && github.event_name != 'pull_request' }}
```
ajbozarth, line 37: As mentioned in my previous comment, this is the line that determines when the workflow should push a release.

Currently this is when the workflow is run on the main branch of the original (non-fork) repo and is not a PR.

@baentsch I was wondering if this should instead be handled as a input instead. Meaning the workflow would only push when run manually with a new `push_release` input set to true. If we still wanted to have it push automatically in some situations we could add other automated triggers for those situations

ajbozarth:
@baentsch another discussion point in addition to  the value of `env.push` (as mentioned above) that will need to be addressed is about the platform images. I originally intended to address #213 in this PR, but while adding push I found a race-condition with the multi-platform images that will need to be decided on.

In order to push a "multi-platform image" the docker build step must be a single build with a list of platforms as a single arg. Essentially we'd have to build all the platforms in one build command, and can't use the matrix to build them in parallel. Trying to include this feature would either make the CI drastically slow and inefficient, or require separate CI for fast PR build & test vs slow multi-platform build & push.

The alternate solution would be to tag each platform image differently, and choose one to also tag latest. This could be done with a few conditionals in the workflow. (we would need to determine the tag to use and how to derive it from the matrix value in the workflow)

There's also the option to only partially address #213 and build/test both platforms but only push the amd64 image. This would require the least edits to the workflows.

If not addressed, the current workflows would push the arm64 image second (since it takes longer to build using the emulator), overwriting the amd64 image

ajbozarth:
> Maybe a similar approach would be helpful here? See [this workflow](https://github.com/open-quantum-safe/ci-containers/blob/7f2a604f783311737cc9ae390746cc626696e037/.github/workflows/push.yml).

I knew there was a way to do this with docker, but I hadn't thought of a way to leverage it in GitHub workflows until looking at yours. Having separate jobs like that would be overkill, but I can add a dependent job (ie a job that waits for the matrix builds to finish) that just runs the docker manifest command to connect the separately tagged images. This would still require deciding how we want to separately tag the platform images though.

baentsch:
> A couple of months back I wrote a workflow to push multiarch images for our CI containers using the ARM runner (so no emulation required). Maybe a similar approach would be helpful here? See [this workflow](https://github.com/open-quantum-safe/ci-containers/blob/7f2a604f783311737cc9ae390746cc626696e037/.github/workflows/push.yml).

_Much_ better: There's hardly any performance/runtime differential between x64 and arm64. Thanks for that pointer @SWilson4 !

> The alternate solution would be to tag each platform image differently, and choose one to also tag latest. This could be done with a few conditionals in the workflow. (we would need to determine the tag to use and how to derive it from the matrix value in the workflow)

That would be sensible.

> There's also the option to only partially address https://github.com/open-quantum-safe/oqs-demos/issues/213 and build/test both platforms but only push the amd64 image. This would require the least edits to the workflows.

That would be less sensible: If we (spend the cycles to) build it, we should also push it.

> If not addressed, the current workflows would push the arm64 image second (since it takes longer to build using the emulator), overwriting the amd64 image

That also would not be sensible (nor necessarily always be the case if we'd be able to use fast arm64 runners): Given the hint by @SWilson4 above, I'm very much against keeping using an emulator (see https://github.com/open-quantum-safe/tsc/issues/5) to begin with.

The current build-in-parallel-and-push-after-jobs-merge workflow in linux.yml already successfully does what we're discussing here, no? See for example https://github.com/open-quantum-safe/oqs-demos/actions/runs/12034148163 

--> Could that approach be used in all jobs (after switching from emulation to real HW)?

baentsch:
PS: Works like a charm: See https://github.com/open-quantum-safe/oqs-demos/actions/runs/12212192366. But does require CI code duplication. Question: Can one specify different runners for different parts of a matrix build?

ajbozarth:
Based on @baentsch comments I realized I missed the real takeaway from @SWilson4 example: the runners. After a bit of research I found why I had not previously considered that solution: GitHub does not currently provide linux/arm64 runners.

After some more research and testing I was able to update the workflows to leverage runners instead of QEMU. This was only possible due to the existence of the `oqs-arm64` self-hosted runner.

I also update the push to tag the images `latest-x86_64` or `latest-arm64` and then created a followup job to create and push a manifest connecting the two images to `latest` (or whatever tag was inputed by a manual call). In this case I was actually able to split out the job into a separate workflow, something that I was unable to do with earlier steps due to minor differences between demos.

### Implementation Note:
For the new push job I wanted the `if` check at the top level so it would skip the job when not pushing rather than "running" the job and not doing anything. The downside is that at the top level the `env` is not accessible, so I had to duplicate the value of `env.push`

### Potential security risks:
While researching runners I ran into the issue that forks do not have access to self-hosted runners, meaning that when the workflows run on a forks main branch they will never start the `oqs-arm64` runner job and *should* timeout after 6 hours. It will still run on a PR though, but the [docs warn of security issues](https://docs.github.com/en/actions/hosting-your-own-runners/managing-self-hosted-runners/managing-access-to-self-hosted-runners-using-groups#creating-a-self-hosted-runner-group-for-an-organization) depending on how the self-hosted runner is set up. Since I have no access to the runner (I only know it exists because other workflows reference it) I can't tell if this is a concern.
It is worth noting that GitHub plans to make arm64 runners [available in Q1](https://github.com/github/roadmap/issues/970), which would replace the self-hosted runner and any potential security issues it has.

### Remaining ToDo
- [ ] Add documentation

### Remaining Open Discussion:
- Is the definition of `env.push` correct? Comment detailing the issue: https://github.com/open-quantum-safe/oqs-demos/pull/321#discussion_r1871744542

ajbozarth:
Small follow up to the above comment:
- Looks like the openvpn test script is a bit finicky and sometime fails, this should be address in a follow up PR since it would need to be fixed in the script not the workflow.
- This PR nearly fixes #284 by adding automatic releases when updates are pushed to `main` and enabling triggering of version releases, but it does not add automated releases of versioned or latest images upon releases of upstream projects (liboqs/oqsprovider). I would suggest that be handled upstream, or manually triggered. I can add #284 to the "Fixes" list if everyone thinks this dresses it enough.

Hawazyn:
Hi @ajbozarth,

Thank you for your patience, and I apologize for the delay. I've just finished the implementation, and you can find the `docker-lint.yml` file [here](https://github.com/Hawazyn/oqs-demos/blob/main/.github/workflows/docker-lint.yml). Please review it and let me know your feedback or suggestions.

ajbozarth:
> I've just finished the implementation, and you can find the docker-lint.yml file [here](https://github.com/Hawazyn/oqs-demos/blob/main/.github/workflows/docker-lint.yml). Please review it and let me know your feedback or suggestions.

@Hawazyn I would suggest opening a PR for review

.github/workflows/openssl3.yml
baentsch: Sounds like a sensible approach to me.

ajbozarth: Which approach are you calling "sensible"? The one I've already coded (# 1) or my suggestion in the comment (# 2)

1. run on the main branch of the original (non-fork) repo and is not a PR
2. only push when run manually with a new push_release input set to true

ajbozarth:
> I'd be OK with the latter if you'd be willing to add code run at/to the upstreams to facilitate that

Once this PR is merged to main I can open PRs on liboqs and oqsprovider adding a small workflow to tigger the build.yml workflow. 

Based on this I will need to update `env.push` to also not run when `input.build_main` is true (which it will be when called from liboqs and oqsprovider.

ajbozarth:
> Once this PR is merged to main I can open PRs on liboqs and oqsprovider adding a small workflow to tigger the build.yml workflow.

I opened https://github.com/open-quantum-safe/oqs-provider/pull/587 as a draft, but for some reason the new workflow didn't run, so I'll have to wait to debug it until this is merged. Once I get that working I will add the same workflow to liboqs

baentsch (commented):
Thanks for agreeing to add release triggers, @ajbozarth . Also very good this is now using the native arm64 runner. The only thing open then from my side is documentation: What's triggered when and how and for which demos? What's a blueprint ("bare minimum CI") for adding new demos? Where and when are (which versions of) the images pushed to?

ajbozarth:
> The only thing open then from my side is documentation: What's triggered when and how and for which demos? What's a blueprint ("bare minimum CI") for adding new demos? Where and when are (which versions of) the images pushed to?

Where do you suggest I add this? Should it go in a new section on the README (if so where in the README) or should I create a new md file?

ajbozarth:
> Where do you suggest I add this? Should it go in a new section on the README (if so where in the README) or should I create a new md file?

@baentsch further followup on my question: I noticed that our contributor docs are in the [GitHub wiki](https://github.com/open-quantum-safe/oqs-demos/wiki/Contributing-guide) and not a markdown file. Is there a motivation for this? The wiki can only be edited with write access and can't take PRs. I think that doc would be the best place to add the CI docs, but I can only contribute it if it's in a markdown file instead of the wiki. Would it be appropriate to move that content into a new CONTRIBUTING.md file?

ajbozarth:
@baentsch So I decided to go through my suggestion above about the CONTRIBUTING.md and move the wiki content to the new file and added a CI section to the bottom. This means we could get rid of the wiki page if you're ok with this change. 

I also updated the README in a few places and added workflow badges to the table for each demo (I left the openssl3 one at the top since it's not in the table?).

Lastly I added a workflow template file for future use. Though I put it in the templates dir I did not add the metadata file to propagate it to the GitHub Action UI. It's just there as a starting point for new demos.

ajbozarth:
#### Update on the liboqs/oqs-provider upstream CI integration

This is much more complicated than I expected. In short there is no way to do what I wanted to, that is to add a workflow to liboqs/oqs-provider that runs the build.yml here and then shows success/failure in the CI for those repos. The method I intended to use try to run the workflow against the repo that runs it, meaning it will run the workflows against liboqs repo contents instead of oqs-demos repo.

The proper way of triggering a workflow from another repo is to make a `repository_dispatch` call. This is done by making a curl POST call from the liboqs workflow that will then trigger a workflow here, but won't track its results or show any helpful feedback in liboqs's CI. In addition `repository_dispatch` works completely different than `workflow_call` or `workflow_dispatch` and has incompatible inputs, meaning I would need to make a lot of edits to build.yml. Also `repository_dispatch` would require setting up a Personal Access Token for a maintainer that has write access to both repos to use in the workflow.

#### In Summary

Based on my research and understanding of what we want from this feature, it would actually be less work and hassle to either just run it manually after updates are made to liboqs/oqs-provider, or set up a cron job to run it (this would still require the edits to build.yml) once a week or something. @baentsch ?

ajbozarth:
Per my above comment I have pushed an update adding a cron job that will run all the workflows against liboqs and oqs-provider `main` once a week (Mon 1am). If we would rather use `repository_dispatch` instead I can revert the last commit and add that instead, but I believe this solution is cleaner and requires less work.

As for displaying the results, IIUC the badges I added to the README will only look at the latest run, so unless a new commit was push in the last week they will show the results of the weekly build against liboqs main. I believe this is an okay behavior (consistent with many other repos). This also means that we can add a badge for build.yml on liboqs and oqs-provider to display that week's results. If this is approved I will open PRs to add those badges after this is merged

baentsch:
> > Where do you suggest I add this? Should it go in a new section on the README (if so where in the README) or should I create a new md file?
> 
> @baentsch further followup on my question: I noticed that our contributor docs are in the [GitHub wiki](https://github.com/open-quantum-safe/oqs-demos/wiki/Contributing-guide) and not a markdown file. Is there a motivation for this? The wiki can only be edited with write access and can't take PRs. I think that doc would be the best place to add the CI docs, but I can only contribute it if it's in a markdown file instead of the wiki. Would it be appropriate to move that content into a new CONTRIBUTING.md file?

This is precisely the same issue/suggestion made https://github.com/open-quantum-safe/oqs-demos/pull/329#issuecomment-2531308429. The reason for not doing it earlier was the simple lack of interest by more people to contribute. This has now changed with you and @Hawazyn being active, so the switch to a file is goodness and follows the same switch we did in `liboqs` and `oqsprovider` (for the same reasons :-). So, Thanks for doing it.

baentsch:
> Based on my research and understanding of what we want from this feature, it would actually be less work and hassle to either just run it manually after updates are made to liboqs/oqs-provider, or set up a cron job to run it (this would still require the edits to build.yml) once a week or something. @baentsch ?

`oqsprovider` has a script to run for validating releases. It may be a good thing to add the above-mentioned "trigger-curl" to that (and add such script to `liboqs` too: @SWilson4 ?). Doing it by "manual-edit-then-cron" (if I understand the approach) sounds not quite right.

Hawazyn:
@ajbozarth This PR is quite extensive, and I appreciate the time and effort you’ve already dedicated to it. While I’m eager to contribute, I feel a bit overwhelmed. If it’s feasible and doesn’t disrupt progress, could you kindly assign me specific tasks along with brief guidance? This would help me support your effort more effectively.

ajbozarth:
> Doing it by "manual-edit-then-cron" (if I understand the approach) sounds not quite right. 

@baentsch I'm not 100% sure we're on the same page here, the cron would just run once a week and try building and testing each demo against liboqs and oqs-provider main. This is more of a regular check, rather than being triggered by a push to main.

If this is still not something we want I can revert https://github.com/open-quantum-safe/oqs-demos/pull/321/commits/666ab72aeebc201f76ae77de0141e85dda0a1642 which contains the cron addition.

ajbozarth:
> Any last comments before this is merged? 

I'd say this is good to merge. 

> Edit/add: Is the openvpn error in CI one worth while investigating?

I believe we will need to open a followup issue to address it, I'm not sure why the test started failing last week. It had previously passed consistently, then began to fail consistently last week. The switch had no relation to the code in the PR afaik, it started after I made a push after not touching the PR for 2 weeks due to my vacation/holidays. Since then it has consistently failed on the ubuntu-latest runner, the arm64 test started passing once I switch from using qemu to the oqs-arm64 runner.

baentsch:
> we will need to open a followup issue to address it

Created #330 to track. Merging this PR then: Thanks for all the effort, @ajbozarth !

Issue #324, opened 2024-11-25
Wireshark docker scan failing


baentsch:
See e.g., https://github.com/open-quantum-safe/oqs-demos/actions/runs/12004752579/job/33460226111

Would you want to take a look @Hayyaaf ?

PR #316, opened 2024-11-14
Update Wireshark Demo

Hayyaaf:
### Summary
This update enhances the Wireshark demo in the `oqs-demos` repository by integrating OpenSSL 3 with the OQS provider for quantum-safe cryptography support. Key improvements include:

- **Updated Dockerfile**:
  - Builds `Wireshark 4.4.1` (latest version) with OpenSSL 3 with the OQS provider.
  - Consolidates the entire build process within the Dockerfile, simplifying configuration and maintenance.

- **Automated Header Generation**:
  - Added `generate_qsc_header.py`, a Python script that automates the generation of `qsc.h`, defining post-quantum cryptographic algorithms for Wireshark.
  - The script fetches algorithm definitions from the `ALGORITHMS.md` file in the OQS provider repository to ensure support for the latest algorithms.
 
- **Project Cleanup and Expanded Documentation**:
  - Expanded `README.md` with comprehensive setup instructions, configuration options, and cross-platform usage guidelines.
  - Removed outdated files (`USAGE.md`, `build.sh`, and `wolfssl-qsc.h`) to streamline the project structure.

### Cross-Platform Notes
Guidance for Linux and macOS display setup was drawn from documentation. Further testing on these platforms is recommended to ensure full functionality and compatibility.

### Testing Summary
The updated Wireshark demo has been verified on **Windows 11** with **Docker 4.35.1** (latest version). Testing confirmed a successful build and functionality of Wireshark with post-quantum cryptography support.

#### Test Screenshots
The following screenshots show the successful integration of post-quantum cryptography in Wireshark. The tests capture traffic using the Kyber1024 and Frodo640aes algorithms.

Screenshot (5)
Screenshot (3)

PR #326, opened 2024-11-26
Follow-up to PR #316: Fix Wireshark Dockerfile and update documentations

Hayyaaf:
### Description

- Removed `OQS_MINIMAL_BUILD` flag from the Dockerfile as it was added mistakenly and caused a bug.
- Updated documentation to reflect confirmed macOS support for the Wireshark demo, as @ajbozarth  indicates it is working.
- Added the `--rm` flag to the testing command for better cleanup after container execution.
- Removed `openssl` and `libssl-dev` from the build stage of the Dockerfile as they are not required for the build process. 

Resolves: #324

### DIFF
--- a/wireshark/Dockerfile
+++ b/wireshark/Dockerfile
@@ -27,7 +27,7 @@ ARG INSTALLDIR
 # Install essential build dependencies
 RUN apt-get update && apt-get install -y --no-install-recommends \
     build-essential libtool automake autoconf cmake ninja-build \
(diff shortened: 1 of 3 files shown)


baentsch (approved):

baentsch:
Merged. Thanks for the improvement @Hayyaaf and the test @ajbozarth !

Hayyaaf:
My pleasure, and thank you all too!

PR #333, opened 2024-12-17
Address issues with the push manifest CI workflow

ajbozarth:
Contrary to my prior understanding, local workflows can not be called in the uses of a step, only in the uses of a job.

I have update all the workflows to use the push-manifest.yml workflow at the job level instead of step.

### DIFF
--- a/.github/workflows/curl.yml
+++ b/.github/workflows/curl.yml
@@ -138,20 +138,26 @@ jobs:
             ghcr.io/${{ github.repository_owner }}/curl:${{ inputs.release_tag || 'latest' }}-${{ matrix.arch }}
             openquantumsafe/curl:${{ inputs.release_tag || 'latest' }}-${{ matrix.arch }}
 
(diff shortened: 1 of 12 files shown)


SWilson4 (approved):

ajbozarth:
As a bit of extra explanation following my comments in todays call: The `uses` of a job can be a workflow as I had already done in `build.yml` but the `uses` of a step must be an "action" not a workflow. Actions work the same way, but are referenced differently. Actions must be located in the `action.yml` and are referenced by their directory, so one action per directory. I could have written the solution by moving push-manifest.yml to an action, but I felt this solution was cleaner and more in line with my other work.

ajbozarth:
Also unlike my previous attempt this was (brute-force) tested on my fork. I deleted/commented out all the code referencing the oqs-arm64 runner and dockerhub so it would successfully run, but the core solution did work.

PR #335, opened 2024-12-18
Update CI to inherit secrets between workflows

### LINKED ISSUE #334: CI fails on auth when trying to push
After my latest fix in #333 The workflow is properly working but it has show a new issue. As detailed[ in the docs here](https://docs.github.com/en/actions/security-for-github-actions/security-guides/using-secrets-in-github-actions#using-secrets-in-a-workflow) secrets are not passed to reusable workflows, meaning when the `push` job calls `push-manifest.yml` that workflow does not have access to the `secrets.DOCKERHUB_USERNAME` and  `secrets.DOCKERHUB_TOKEN` values. Based on this it also means if `build.yml` is run manually on `main` it would also fail for the same reason.

On preliminary research this should be able to be addressed by follow [this doc](https://docs.github.com/en/actions/sharing-automations/reusing-workflows#using-inputs-and-secrets-in-a-reusable-workflow).

ajbozarth:
By default called workflows don't have access to secrets, this causes issues when needing access to our dockerhib secrets for push.

This adds `secrets: inherit` to all local workflow calls.

Fixes #334

### DIFF
--- a/.github/workflow-templates/template.yml
+++ b/.github/workflow-templates/template.yml
@@ -92,9 +92,8 @@ jobs:
   push:
     if: ${{ github.repository == 'open-quantum-safe/oqs-demos' && github.ref == 'refs/heads/main' && github.event_name != 'pull_request'  && inputs.build_main != 'true' }}
     needs: build
-    runs-on: ubuntu-latest
-    steps:
-      - uses: ./.github/workflows/manifest.yml
-        with:
-          image_name: <demo_name>
-          release_tag: ${{ inputs.release_tag || 'latest' }}
+    uses: ./.github/workflows/push-manifest.yml
+    secrets: inherit
+    with:
+      image_name: <demo_name>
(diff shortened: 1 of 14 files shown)


ajbozarth:
Sorry for the continued CI thrash, this was another case of being unable to test prior to merge since it's an issue with secrets. Note that is still the case with this fix, I have no way to test it prior to merge, but based on the following excepts from [this docs section](https://docs.github.com/en/actions/sharing-automations/reusing-workflows#creating-a-reusable-workflow) I believe this should be the fix.

> Workflows that call reusable workflows in the same organization or enterprise can use the `inherit` keyword to implicitly pass the secrets.

> If the secrets are inherited by using `secrets: inherit` in the calling workflow, you can reference them even if they are not explicitly defined in the `on` key.

baentsch (approved):
let's give it a try

ajbozarth:
And finally success! All green on main (except the failing openvpn one)

Issue #300, opened 2024-09-25
Consider use of opencontainers spec metadata in container images


planetf1:
The opencontainers spec defines a list of standard annotations which could be used to add in important metadata. This superceeds the earlier label schema

See https://github.com/opencontainers/image-spec/blob/main/annotations.md

For instance this might be
```
LABEL org.opencontainers.image.version="5"
```

It may be worth opening a new issue just on this specific aspect

_Originally posted by @planetf1 in https://github.com/open-quantum-safe/oqs-demos/pull/298#discussion_r1775072942_

hawazyn:
@baentsch @ajbozarth  I would like to take this issue. However, there are a lot of annotations in the OCI specification. Which of these should I prioritize, and are there any that should be left out for now? Looking forward to your guidance

hawazyn:
@baentsch @ajbozarth  I would like to take this issue. However, there are a lot of annotations in the OCI specification. Which of these should I prioritize, and are there any that should be left out for now? Looking forward to your guidance.

dstebila:
Closing as not planned.

PR #348, opened 2025-02-09
Remove GitHub Actions workflow for Docker image scanning

hawazyn:


SWilson4:
Hi @hawazyn—what's the rationale for deleting the workflow?

hawazyn:
Hi @SWilson4, The rationale for deleting the workflow is that it’s not scanning all images, leading to pipeline failures. Additionally, there’s redundancy in the process since we build the image and then scan it, whereas with Alex CI pipeline changes, we can scan images directly, simplifying the workflow and reducing overhead.

SWilson4:
> Hi @SWilson4, The rationale for deleting the workflow is that it’s not scanning all images, leading to pipeline failures. Additionally, there’s redundancy in the process since we build the image and then scan it, whereas with Alex CI pipeline changes, we can scan images directly, simplifying the workflow and reducing overhead.

OK—but are we scanning the images in other workflows currently? If not, I would prefer not to remove this workflow until the scans are happening elsewhere in CI.

hawazyn:
I’m currently reviewing the pipelines to determine the best way to implement image scanning. I may also consider addressing #300 if it aligns with the updates. Let me know if you have any guidance on these!

hawazyn:
I was planning to do it later, but I changed my mind and decided to work on it now.

hawazyn:
@baentsch  @SWilson4 

The workflows have been updated and tested on several demos, selected randomly. Due to the GitHub Actions limitation:

“too many workflows are referenced, total: 21, limit: 20,”

I couldn’t test all of them. Full functionality can be verified after integration. I kindly request your review.

hawazyn:
@baentsch  @SWilson4 

The workflows have been updated and tested on several demos, selected randomly. Due to several limitation. I couldn’t test all of them. Full functionality can be verified after integration. Additionally, the CI process is time-consuming due to Docker builds.

Special thanks to @ajbozarth and anyone else who contributed to writing and managing this complex workflow and for all the hard work put into it. I kindly request your review.

baentsch:
I wonder why all CI jobs are still failing (on login to docker hub -- I thought this step has been eliminated @ajbozarth ? Can you shed some light on this?)...

.github/workflow-templates/template.yml
```diff
@@ -48,14 +65,18 @@ jobs:
...
             runner: ubuntu-24.04-arm
+
     runs-on: ${{ matrix.runner }}
+
     steps:
       - uses: actions/checkout@v4
+
+      # Log in to Docker Hub is required for scanning and pushing Docker images
       - uses: docker/login-action@v3
-        if: env.push == 'true'
```
SWilson4, line 55: This line is essential to ensure that logins aren't attempted at every PR.

SWilson4 (dismissed):

.github/workflow-templates/template.yml
hawazyn: I removed it because it’s required—we need to log in to Docker to use Docker Scout. Should I adjust it so that the scan is triggered only during a push to Docker Hub?

hawazyn: I think I’ll bring this condition back. I’ll also add it to the Docker scan and upload steps so the CI doesn’t fail unnecessarily. This way, the scan will only run when pushing to Docker Hub. Let me know if you agree with this approach.

SWilson4: That seems reasonable to me: it looks like scanning previously only ran on commits to main, which also involved pushes.

baentsch (approved):

baentsch:
@hawazyn This LGTM. Will merge on Monday if @SWilson4 doesn't object.

SWilson4:
@hawazyn Looks like there were some failures on main after merge. Could you take a look at them?

hawazyn:
This [workflow](https://github.com/open-quantum-safe/oqs-demos/actions/runs/13421069421/job/37493585080) failed due to a build error, whereas the other failures were caused by Docker Scout. Since the build was successful before merging, re-running the workflow should likely resolve the issue.

hawazyn:
Good evening, @SWilson4. This [workflow](https://github.com/open-quantum-safe/oqs-demos/actions/runs/13421069421/job/37493585080) failed due to a build error, while the other failures were caused by Docker Scout. Since the build was successful before merging, re-running the workflow should likely resolve the issue.

hawazyn:
The CI failures have been successfully resolved.

baentsch:
@hawazyn Thanks again for the latest updates which got the main project to an overall "green" CI status -- it seems the curl project isn't quite right, though: It's showing a CI failure for `aarch64`: This seems to have been introduced by this PR, so could you please check if that is the case? If the error has another cause, please raise an issue for someone (most likely me :) to look into fixing it? Thanks in advance.

hawazyn:
Thanks for checking! I just took a look, and the CI is green now. It seems the issue has already been resolved. Let me know if you still see any concerns!

hawazyn:
@baentsch Thanks for checking! I just took a look, and the CI is green now. It seems the issue has already been resolved. Let me know if you still see any concerns!

baentsch:
> I just took a look, and the CI is green now. It seems the issue has already been resolved. Let me know if you still see any concerns!

Hmm -- I see this (curl failing):
grafik

But then again, it seems this is an infrastructure problem (arm not completing/running extremely slow: Just triggered a re-run to observe). @ryjones : Are you aware of problems due to lack of CI funding or such (altghough I thought no payment is necessary for arm any more)? If technical infrastructure problem, already asked at https://github.com/orgs/community/discussions/148648#discussioncomment-12297086.

hawazyn:
> But then again, it seems this is an infrastructure problem (arm not completing/running extremely slow: Just triggered a re-run to observe). @ryjones : Are you aware of problems due to lack of CI funding or such (altghough I thought no payment is necessary for arm any more)? If technical infrastructure problem, already asked at https://github.com/orgs/community/discussions/148648#discussioncomment-12297086.

The issue is GitHub Actions’ 20 concurrent workflow limit, but the current workflow triggers 21 jobs, causing failures ([GitHub Docs - Workflow Limitations](https://docs.github.com/en/enterprise-cloud@latest/actions/sharing-automations/reusing-workflows#limitations)).

Best Solutions in the Current Situation:
1. Use Multi-Arch Docker Build via Buildx – Instead of running separate jobs for ARM and x86_64, combine them into one job using Docker Buildx. This reduces the job count while maintaining multi-arch support.
2. Split Jobs into Two Workflows – Separate ARM and x86_64 builds into two workflows to stay within the limit.

hawazyn:
> But then again, it seems this is an infrastructure problem (arm not completing/running extremely slow: Just triggered a re-run to observe). @ryjones : Are you aware of problems due to lack of CI funding or such (altghough I thought no payment is necessary for arm any more)? If technical infrastructure problem, already asked at https://github.com/orgs/community/discussions/148648#discussioncomment-12297086.

The issue is GitHub Actions’ 20 concurrent workflow limit, but the current workflow triggers 21 jobs, causing failures ([GitHub Docs - Workflow Limitations](https://docs.github.com/en/enterprise-cloud@latest/actions/sharing-automations/reusing-workflows#limitations)).

I think the best solution in the current situation is to use Multi-Arch Docker Build via Buildx. Instead of running separate jobs for ARM and x86_64, combine them into one job using Docker Buildx. This reduces the job count while maintaining multi-arch support.

baentsch:
> I think the best solution in the current situation is to use Multi-Arch Docker Build via Buildx. 

I'm not sure of that: Do you know how this is implemented? I'm a bit afraid that it'd be using cross-compilation for other platforms or in the worst case, emulation -- making the build _really_ slow. [The documentation](https://docs.docker.com/build/building/multi-platform/) hints at all options to be supported, so the question is what is done in GH by default...

hawazyn:
> I'm not sure of that: Do you know how this is implemented? I'm a bit afraid that it'd be using cross-compilation for other platforms or in the worst case, emulation -- making the build _really_ slow. [The documentation](https://docs.docker.com/build/building/multi-platform/) hints at all options to be supported, so the question is what is done in GH by default...

I tried using Buildx, and it was painfully slow, unbearable. I think we should separate the ARM and x86_64 build calls, ensuring each is triggered independently.

PR #361, opened 2025-02-28
Updated oqs and provider version in readme

christianpaquin:
Updated the liboqs and oqs-provider version in the readme, to match what is documented in [https://test.openquantumsafe.org/](https://test.openquantumsafe.org/)


baentsch (approved):
Thanks for the catch @christianpaquin ! I'm really afraid we're more and more losing the hang of this.

PR #369, opened 2025-05-07
Re-add epiphany demo

davidkel:
This has some caveats to be aware of
1. No idea if it would work on apple or windows OS as they are not linux and would also need an X Server
2. Image is fixed to Ubuntu 22.04. Can't update the base image it won't work.
3. Tested on a Ubuntu 24.04 VM and Raspian Bookworm 64bit on a raspberry pi.

In general running GUIs from containers is complex requiring various other components such as dbus to be running. In the case here it manages to work with an environment variable hack, but moving to 24.04 introduced a new issue with dbus. KDE apps may also have issues. Even X Can have issues like the shared memory extension. There are interesting tools such as X11Docker which try to provide easier solutions but on the whole trying to run gui applications from a container can be fragile.


davidkel:
looks like arm failed with a timeout trying to get the oqs certificate, force pushed to do a rebuild

davidkel:
npw looks like the arm build has hung in the middle of building openssl, so seems there are issues with the github arm runners at the moment

bhess:
Thanks for this work, David!

davidkel:
the github actions issue with arm appears to be resolved and the arm build has now worked

epiphany/USAGE.md
```diff
@@ -4,34 +4,30 @@ This docker image contains a version of the [GNOME Web/epiphany](https://github.
 
 To this end, it contains QSC algorithms implemented by [liboqs](https://github.com/open-quantum-safe/liboqs) and made available to OpenSSL(3) via [oqs-provider](https://github.com/open-quantum-safe/oqs-provider) developed as part of the [OpenQuantumSafe](https://openquantumsafe.org) project.
 
-The image is based on Ubuntu and requires the host to run the Unix X-Window system.
-
-This demo is based on work done by [Igor Barshteyn](https://www.linkedin.com/pulse/demonstrating-quantum-safe-tls-13-web-server-client-nist-barshteyn).
```
SWilson4, line 9: Is this demo still based on Igor's work? I feel like we shouldn't drop the acknowledgement completely.

davidkel: Not that I am aware of as I've not looked at the work. I did start out by looking at what was in the oqs-demos version already, but in the end the Dockerfile is based on the format similar to node.js ie the building of openssl and oqs (although improved so that there is no dependency on openssl in the repo now as there is in the others which we could update the other demos to replicate). The building of glib-networking is updated slightly from what was present in the current demo, but the work to piece it all together, get it to run and comment as to why the various env vars were needed is all my own work and the information to get it to work from a container image was refined from what I found. So I don't feel it's really based on work done by the link reference anymore (I had already determined that epiphany didn't use openssl and actually used gnuTls via the glib-network module based on something else I was trying) but of course that was present in the currently not working demo.
I can re-add the link though if you feel it should be there.

SWilson4: OK, makes sense. In that case, I think it is OK to drop the link from this documentation: Igor is acknowledged in the README contributor list and on [openquantumsafe.org](https://openquantumsafe.org/applications/external.html), and linking to a discussion/tutorial on a previous version here may confuse users.

SWilson4 (approved):

epiphany/Dockerfile
```diff
@@ -0,0 +1,127 @@
+
+# Define build arguments for version tags, installation paths, and configurations
+ARG UBUNTU_VERSION=22.04 # does not work on 24.04 `Failed to create XdpPortal instance: Failed to execute child process "dbus-launch" (No such file or directory)`
+ARG INSTALLDIR=/opt/oqssa
+
+# OpenSSL/OQS Versions
+ARG OPENSSL_TAG=openssl-3.4.0
```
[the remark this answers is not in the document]
davidkel: @bhess We could use those versions, a couple of thoughts though are being consistent with the rest of the demos wrt to versions makes it easier to do a global change to move them all to the next versions for the demos to be based off of (ideally what would be good would be an easy way for all demos to work off a single place where the openssl and oqs versions are defined for all demos)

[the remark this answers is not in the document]
bhess: I agree that basing them on releases is preferable. Having a central configuration for versioning would be nice to keep everything in sync. I'll leave the decision to you whether to pin the new liboqs now or wait until the oqs-provider has been released.

bhess (approved):
LGTM

SWilson4:
Merging as all outstanding comments have been resolved. Thanks @davidkel!

PR #372, opened 2025-06-10
Update various demos

praveksharma:
This updates the following demos to liboqs 0.13.0/oqs-provider 0.9.0:

 - curl
 - ephiphany
 - h2load
 - haproxy
 - httpd
 - mosquitto
 - nginx
 - openvpn
 - openssh
 - nodejs


epiphany/Dockerfile
```diff
@@ -5,8 +5,8 @@ ARG INSTALLDIR=/opt/oqssa
 
 # OpenSSL/OQS Versions
 ARG OPENSSL_TAG=openssl-3.4.0
```
davidkel, line 7: I think we should update to openssl-3.5.0

davidkel (changes requested):
I think the Label version for node and epiphany need to be updated as well

epiphany/Dockerfile
praveksharma: This was a deliberate choice because OpenSSL 3.5.0 comes with its PQ implementations, so if we use OpenSSL 3.5.0 liboqs implementations of ML-KEM and ML-DSA would not be used in demos. What are your thoughts?

praveksharma:
Marking as draft while I resolve CI errors.

epiphany/Dockerfile
davidkel: As OpenSSL 3.5 is the official LTS release I thought it would be better to use it. OQS supplements what is already present in OpenSSL 3.5 (although it looks like this combination still doesn't allow hybrid certs so that maybe a good reason of course). Happy to go with what the main maintainers think

PR #380, opened 2025-08-29
Add plotly digital signatures visualization demo

iyanmv:
This PR adds a Plotly-based webapp to visualize the characteristics (keys and certificate sizes) and the performance (keygen, sign, verify) of the digital signatures supported by liboqs as well as common traditional signatures (for this [PyCryptodome](https://www.pycryptodome.org/) is used).

A public instance running the demo is available [here](https://pqc.crypto-lab.ch/sig-charts/) and its source code is also available [here](https://github.com/cryptohslu/pqc-sig-charts-website). The code in this PR removes all the branding (favicon, links, etc.).

This goal of this visualization is different from the benchmarking plots available in https://openquantumsafe.org/benchmarking/ and it's targeting more general audience.

I can maintain this demo going forward, and I could also help adapt the code to something that `liboqs` could run in parallel to the performance visualizations.

@SWilson4 @pablo-gf This is the webapp we discussed some months ago. Perhaps you can have a look and give one first round of feedback :) I will add the CI workflow file and modify the README as well in the next days.

## Things left to do
- [x] Add the CI workflow
- [ ] Add tests to workflow (any ideas how?)
- [ ] Add to README


dstebila:
That's really cool looking! I think it provides a helpful high-level comparison between the different algorithms.

Are the CI errors blocking this something that you need help with resolving from our side or something that you're still working through?  And once it's ready here, then we would also want to integrate it into the website via the https://github.com/open-quantum-safe/www repository.

baentsch (changes requested):
Thanks @iyanmv for this work: It's a nice, lightweight way to show & compare the "topline" properties of all, incl. classic algorithms. Besides the single comments, there's one very basic question: How does this capture (at the very least document) the platform these measurements have been taken on? Or taking it a step further, would it be possible to extend this to extend across platforms and/or across build (incl. optimization) options? Without that, I'm not sure whether this is an apples-to-apples comparison wrt classic crypto and whether one doesn't have to do a lot of explaining how relevant this is (and where).
Again, don't get me wrong: I like this to get a quick overview and the visualization and comparison feature is truly good. But some way to document the platform the numbers have been collected on should be added at minimum. Next stop: Understanding why this still doesn't pass CI after I pulled the code into the OQS namespace -- stay tuned... but please feel free to address these comments already.

iyanmv:
Hi @baentsch! I've been slowly improving this in the last months. I think now I'm ready to push all the changes back here. I think I have addressed most of your comments. I'll ping you when it's ready for a second round of review. Since there are many changes, should I close this PR and start from scratch, or do you prefer me to modify this branch?

Regarding this:
> Or taking it a step further, would it be possible to extend this to extend across platforms and/or across build (incl. optimization) options? Without that, I'm not sure whether this is an apples-to-apples comparison wrt classic crypto and whether one doesn't have to do a lot of explaining how relevant this is (and where).

I have done the following changes:

- Do not use PyCryptodome to benchmark the traditional signatures. This was actually an unfair comparison. For example, we were seeing that the verification of ML-DSA-44 was faster than RSA-2048. I moved all the benchmark code outside the visualization [here](https://github.com/cryptohslu/pqc-sig-bench-c), and now we use liboqs and OpenSSL directly. Hopefully a bit more fair comparison. Also, I think it's better to decouple the benchmarking from the visualization. It should be possible to just put a different dataset here obtain from a different benchmarking run.
- Before I was just running on my computer and that's it. Now I have rerun all the tests in different environments and platforms (AWS EC2 instances, RPi, MacBook, etc.). The dataset can be changed in the visualization, and it is even possible to compare the performance between different datasets. I have not run in the same platform with different optimizations, but this could be added in the future (did you have something in mind?). In any case, I will document exactly how liboqs and OpenSSL were built (the script is also available [here](https://github.com/cryptohslu/pqc-sig-bench-c/blob/main/scripts/run_aws.sh))
- Now there is an "intro tour" explaining how to use and navigate through the visualization. This is only shown the first time someone enter the website, but it can be trigger again clicking on the ? icon in the header bar.
- Many improvements here and there from feedback I got from people playing with the visualization (the format of the comparison table, the interface to select algorithms in the overview page, etc., etc.)

All these changes are already live [here](https://pqc.crypto-lab.ch/sig-charts/).

baentsch:
Thanks for this update @iyanmv ! Based on the explanations above, I think a new PR is warranted. This is new --and exciting-- stuff: I just looked at your live system and I like it very much. With one caveat: You state "The smaller the enclosed area, the better": Isn't this counter-intuitive for radar charts? Isn't it normal that better is _more_ area enclosed? But then again, radar charts can be misleading anyway already by setting up the axes, no (maybe :) don't mind...

PR #382, opened 2025-10-06
Update nginx-quic.conf

pi-314159:
Replace `mlkem1024` with `p384_mlkem1024`, as the former was using Google's implementation.

### DIFF
--- a/nginx/nginx-conf/nginx-quic.conf
+++ b/nginx/nginx-conf/nginx-quic.conf
@@ -29,7 +29,7 @@ http {
         ssl_certificate		/certs/server.crt;
         ssl_certificate_key	/certs/server.key;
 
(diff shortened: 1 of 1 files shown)


baentsch:
> Replace `mlkem1024` with `p384_mlkem1024`, as the former was using Google's implementation.

OK..... Does that mean google's mlkem1024 doesn't interoperate with the OQS one? If so, any idea why? Just changing the algorithm used for testing may "turn CI green" but just "plaster over" a real problem, no?

pi-314159:
> OK..... Does that mean google's mlkem1024 doesn't interoperate with the OQS one? If so, any idea why? Just changing the algorithm used for testing may "turn CI green" but just "plaster over" a real problem, no?

They work well. The issue was with the Nginx configuration. Our implementation uses `mlkem1024`, whereas Google uses `MLKEM1024`, the official name. This mismatch caused the nginx error

PR #394, opened 2026-05-17
Fix CI tags in README and disable unmaintained projects in build.yml

danardiello:
In the README each sub-project badge pointed to a separate and sometimes outdated workflow run, which caused them to still show "passing" even when the sub-project was failing in the main build run. Since workflow status badges don't support adding a job parameter (only branch and event), i displayed the overall build status instead and added a note to check the workflow link for detailed job results. If "per-project badges" are still desired, a separate workflow might be required for this.

I also excluded all projects marked as unmaintained in the README from the build workflow to avoid wasting CI resources.


baentsch (approved):
LGTM, particularly saving on resources regarding unmaintained stuff. Thanks @danardiello !

PR #397, opened 2026-06-13
Update OQS dependencies and fix segmentation faults on Alpine

Keelan10:
This PR updates the h2load demo environment to newer OQS component versions and fixes BIKE-related segmentation faults observed on Alpine.


baentsch:
Thanks for this update, @Keelan10 . By introducing OpenSSL 3.6, though, this also adds PQ code from the default provider. This begs the question whether this build now should also have some basic functionality tests, maybe showing exactly which PQ code actually is executed... Possible? Sensible?

Issue #398, opened 2026-06-21
[New Demos] PostgreSQL, gRPC, and Apache Kafka PQC TLS integration demos


netanmangal:
## Summary

I'd like to contribute new integration demos for infrastructure categories that are currently unrepresented in this repository: **databases**, **RPC frameworks**, and **streaming platforms**.

The existing 14 demos cover web servers, browsers, CLI tools, VPN/SSH, MQTT, and proxies but nothing for the data infrastructure layer that enterprises need to migrate for PQC compliance.

### 1. PostgreSQL + PQC TLS
- PostgreSQL 18 with `ssl_groups = X25519MLKEM768:X25519:prime256v1`
- Server compiled against OpenSSL 3.5+ (or using oqs-provider)
- Client connection via libpq with PQC key exchange
- Docker Compose setup
- **Why:** PostgreSQL is the world's most popular open-source RDBMS. PQC TLS works
  with its `ssl_groups` parameter but no demo exists anywhere.

### 2. gRPC + PQC TLS
- gRPC server/client with hybrid PQC TLS (X25519MLKEM768)
- Python and/or Go examples using OpenSSL 3.5 or BoringSSL
- Performance benchmarking script comparing classical vs PQC handshakes
- **Why:** gRPC merged PQC as default for BoringSSL builds (June 2026), but no
  standalone demo shows how to configure it with oqs-provider.

### 3. Apache Kafka + PQC TLS
- Kafka broker with PQC TLS via Bouncy Castle BCJSSE provider (JDK 24+)
- Producer/consumer connecting with hybrid X25519MLKEM768
- Docker Compose with ZooKeeper/KRaft
- **Why:** Kafka has zero PQC support or documentation (I've filed a JIRA ticket
  to track this). A demo would be the first PQC reference for the Kafka ecosystem.

## Approach

Each demo will follow the existing repo pattern:
- `Dockerfile` with OpenSSL 3.5 or oqs-provider compiled in
- `docker-compose.yml` for easy testing
- `README.md` with step-by-step instructions and verification steps
- CI workflow (GitHub Actions) per CONTRIBUTING.md
- Maintenance commitment

I'll start with **PostgreSQL** (most straightforward - it uses OpenSSL directly) and iterate from there.

## Regulatory Context

[...]

PR #399, opened 2026-06-22
Add PostgreSQL PQC TLS demo

netanmangal:
Add PostgreSQL PQC TLS demo. It's the first database integration in this repository.

Relates to #398

**What's included:**
- `postgresql/Dockerfile` — Multi-stage Alpine build: OpenSSL 3.4 + liboqs 0.13.0 + oqs-provider 0.9.0 + PostgreSQL 17.2. Generates ML-DSA-65 CA and server certificates. Switchable OpenSSL version via `OPENSSL_TAG` build arg.
- `postgresql/docker-entrypoint.sh` — Initializes PGDATA with `scram-sha-256` auth and SSL config.
- `postgresql/docker-compose.yml` — Localhost-bound, configurable `DEFAULT_GROUPS`.
- `postgresql/test.sh` — Verifies SSL enabled, TLS 1.3, ML-DSA-65 cert, PQC peer signature.
- `postgresql/README.md` + `USAGE.md` — Build options, testing instructions.
- Root `README.md` updated with PostgreSQL in the Maintained table.

**Security:** CA key deleted before runtime stage. Auth is `scram-sha-256`, not `trust`. Port bound to `127.0.0.1`.

**Note:** PostgreSQL 17's `ssl_ecdh_curve` only supports classical ECDH. PQC KEM requires PostgreSQL 18's `ssl_groups`. This demo demonstrates PQC *authentication* (ML-DSA-65 certificates).

Docker builds, all 4 tests pass.


postgresql/README.md
```diff
@@ -0,0 +1,100 @@
+## Purpose
+
+This directory contains a Dockerfile that builds [PostgreSQL](https://www.postgresql.org/) using OpenSSL with the [OQS provider](https://github.com/open-quantum-safe/oqs-provider), which allows PostgreSQL to use quantum-safe authentication (PQC certificates) in TLS 1.3 connections between clients and the database server.
+
+The demo generates CA and server certificates signed with ML-DSA-65 (a NIST-standardized post-quantum signature algorithm), ensuring that TLS authentication is quantum-resistant. The key exchange uses classical ECDH as PostgreSQL 17's `ssl_ecdh_curve` parameter does not yet support PQC KEM groups (PQC KEM support requires PostgreSQL 18+ with the `ssl_groups` parameter).
```
baentsch: That begs the obvious question, why this build does not use 18+?

postgresql/Dockerfile
```diff
@@ -0,0 +1,157 @@
+# Define build arguments for version tags, installation paths, and configurations
+ARG ALPINE_VERSION=3.21
+ARG OPENSSL_TAG=openssl-3.4.0
+ARG LIBOQS_TAG=0.13.0
+ARG OQSPROVIDER_TAG=0.9.0
+ARG POSTGRESQL_VERSION=17.2
```
baentsch: This seems already perfectly parameterized, so why not go to 18 if this enables PQ-KEMs (?)

postgresql/README.md
netanmangal: You're right.

postgresql/Dockerfile
netanmangal: Well, you're right.

image


I did a very quick search over Google to see the most stable version and Gemini gave me 17. I went onto working with 17 without a second thought 😆 😆

postgresql/README.md
netanmangal: I will update the PR.

postgresql/docker-entrypoint.sh
```diff
@@ -0,0 +1,53 @@
...
+EOF
+
+    echo "PostgreSQL initialized with PQC TLS configuration."
+    echo "Signature algorithm: ${SIG_ALG:-mldsa65}"
+    echo "KEM groups: ${DEFAULT_GROUPS}"
+
+    # Verify OpenSSL providers are loaded
+    echo "OpenSSL version and providers:"
+    openssl version
+    openssl list -providers 2>/dev/null || true
```
baentsch, line 50: Would it be better to check _which_providers are loaded? In general the question: How do you verify that indeed PQ auth/KEM (if possible) is active?

postgresql/docker-entrypoint.sh
```diff
@@ -0,0 +1,53 @@
...
+ssl_key_file = '/opt/pqcerts/server.key'
+ssl_ca_file = '/opt/pqcerts/CA.crt'
+listen_addresses = '*'
+EOF
+
+    # PostgreSQL 18+ supports ssl_groups for PQC KEM key exchange
+    PG_MAJOR=$(postgres --version | sed 's/.*) //' | cut -d. -f1)
+    if [ "${PG_MAJOR}" -ge 18 ] 2>/dev/null; then
+        echo "# PQC KEM key exchange (PostgreSQL 18+)" >> "${PGDATA}/postgresql.conf"
+        echo "ssl_groups = 'X25519MLKEM768:X25519:prime256v1:secp384r1'" >> "${PGDATA}/postgresql.conf"
```
baentsch: This groups list seems to contradict/be different from the default groups list in README and Dockerfile. Why?

postgresql/test.sh
```diff
@@ -0,0 +1,138 @@
...
+    if [ -n "${SSL_GROUPS}" ]; then
+        echo "PASS: ssl_groups configured: ${SSL_GROUPS}"
+    else
+        echo "WARN: Could not verify ssl_groups setting"
+    fi
+
+    # Perform a TLS handshake requesting X25519MLKEM768 and compare
+    # the client key_share size against a classical-only handshake.
+    # X25519MLKEM768 key shares are ~1200 bytes; X25519 is 32 bytes,
+    # so the "written" byte count is a reliable differentiator.
```
baentsch: well this looks like a pretty "heuristic" check; while I'm happy to see this exists (answering my question from earlier :), couldn't one assume that by requesting _only_ MLKEM768_X25519 no connection should be established if the server doesn't support it? Could it really fall back to X25519??

baentsch (commented):
Basically, LGTM: Nice work @netanmangal ! See the single comments and please also consider adding it to the CI chain (see .github/workflow files) so that an image directly gets built-and-tested for easy consumption.

postgresql/test.sh
netanmangal: Uhm, right! If we request only X25519MLKEM768, the connection should fail if the server doesn't support it. 

Replaced the byte-size heuristic with a direct test:                                                                      
```
  echo | openssl s_client -connect 127.0.0.1:${PG_PORT} \                                                                   
      -groups X25519MLKEM768 -tls1_3 2>&1                                                                                   
```
                                                                                                                       
If the TLS 1.3 handshake succeeds with only that group offered, PQC KEM is confirmed. If it fails, the test reports FAIL.

postgresql/docker-entrypoint.sh
netanmangal: Updated. the entrypoint now uses ssl_groups = '${DEFAULT_GROUPS}' so it references the same env var set in the Dockerfile.

postgresql/docker-entrypoint.sh
netanmangal: Good point. I've updated the entrypoint to parse `openssl list -providers` output and explicitly check for the OQS provider.

For verifying PQ auth is active, the test script now checks Peer signature type: mldsa65 in the openssl s_client handshake output. For PQC KEM, it attempts a connection with  -groups X25519MLKEM768 exclusively — if the handshake succeeds, PQC KEM is confirmed (similar to comment 3).

netanmangal:
Hi @baentsch ,

Cna you please merge this PR if all looks good. I'll get started with the next one.

baentsch:
> Hi @baentsch ,
> 
> Cna you please merge this PR if all looks good. I'll get started with the next one.

Sure thing. Please look at CI, though: It seems the Arm build failed.

netanmangal:
@baentsch , I have pushed in the fix. Can you see if all is good now to merge. Thanks!

postgresql/docker-entrypoint.sh
```diff
@@ -0,0 +1,57 @@
...
+        echo "PostgreSQL ${PG_MAJOR} detected: ssl_groups not available (requires PostgreSQL 18+)."
+        echo "PQC authentication is active, but key exchange uses classical ECDH."
+    fi
+
+    # Configure client authentication - require SSL for remote connections
+    # NOTE: This is a demo configuration. For production use, review and harden these settings.
+    cat > "${PGDATA}/pg_hba.conf" <<EOF
+# TYPE  DATABASE        USER            ADDRESS                 METHOD
+# Local connections (password not required for local socket connections in this demo)
+local   all             all                                     scram-sha-256
```
baentsch: Why no SSL also locally?

postgresql/docker-entrypoint.sh
```diff
@@ -0,0 +1,57 @@
...
+    echo "Signature algorithm: ${SIG_ALG:-mldsa65}"
+    echo "KEM groups: ${DEFAULT_GROUPS}"
+
+    # Verify OpenSSL providers are loaded
+    echo "OpenSSL version and providers:"
+    openssl version
+    if openssl list -providers 2>/dev/null | grep -qi "oqs"; then
+        echo "OQS provider: loaded"
+    else
+        echo "WARNING: OQS provider not detected"
```
baentsch: Why this warning? As long as PQ works (it also may by way of `openssl` default provider), all is fine, right?

postgresql/README.md
```diff
@@ -0,0 +1,113 @@
...
+
+Complete information on how to use the image is [available in the separate file USAGE.md](USAGE.md).
+
+## Build options
+
+The Dockerfile provided allows for significant customization of the image built:
+
+### OPENSSL_TAG
+
+Tag of `openssl` release to be used. Default is `openssl-3.4.0`.
```
baentsch, line 34: Suggest adding a statement what happens when using a more current `openssl` version (everything with standardized PQ should work identically, just using default provider; non-standard PQ still using/working via `oqsprovider`).

postgresql/README.md
```diff
@@ -0,0 +1,113 @@
...
+To connect with a PQC-enabled `psql` client, you can use the OQS-enabled OpenSSL build. For testing from the container itself:
+
+```
+docker exec -it oqs-postgresql psql -U postgres -h 127.0.0.1 "sslmode=require"
+```
+
+To verify the PQC TLS handshake with `openssl s_client`:
+
+```
+docker exec -it oqs-postgresql openssl s_client -connect localhost:5432 -starttls postgres
```
baentsch: This is a localhost test which seems to have been configured to allow no TLS (see previous comment)

postgresql/test.sh
```diff
@@ -0,0 +1,139 @@
...
+    if [ "$i" -eq 30 ]; then
+        echo "ERROR: PostgreSQL failed to start within 30 seconds."
+        docker logs "${CONTAINER_NAME}"
+        docker rm -f "${CONTAINER_NAME}" 2>/dev/null
+        exit 1
+    fi
+    sleep 1
+done
+
+# Verify SSL is enabled (via local socket, which does not require TLS)
```
baentsch: I don't understand this comment (why via local socket? Why without requiring TLS? It should show that it's using SSL, no?)

postgresql/USAGE.md
```diff
@@ -0,0 +1,83 @@
...
+
+```
+docker exec -it oqs-postgresql psql -U postgres "sslmode=require"
+```
+
+## Advanced usage options
+
+### Configuring KEM algorithms
+
+This PostgreSQL image supports all quantum-safe key exchange algorithms [presently supported by oqs-provider](https://github.com/open-quantum-safe/oqs-provider#algorithms). You can control which algorithms are available via the `DEFAULT_GROUPS` environment variable:
```
baentsch: Not under all circumstances: I'd hope that everything also works fine with some PQ algs using `openssl` >= 3.5.

baentsch (approved):
Thanks for the various improvements @netanmangal ! We could merge as-is, but I found further nits (see single comments): My personal preference would be to see them addressed before a merge in the interest of improved user information, but I'd not be hung up on this: This PR already is better than many other demos we have.

netanmangal:
Perfecto. I look forward to closing, and this PR merged. I'll be happy to create next demo. Thanks!

baentsch:
Merged: Thanks for the contribution @netanmangal !

